PR #7, opened 2021-12-22
Update overview.md

GrafanaWriter:
Minor edits to smooth. I also removed the capitalization of terms. It will be confusing to maintain it here, but not in other content. And the terms are well defined, so they do not need to remain capitalized.

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -21,7 +21,7 @@ Thema is built from six fundamental logical components:
 * **Lens:** A bidirectional mapping between the final schema in one sequence, and the first schema in another sequence.
 * **Lineage:** An ordered list of sequences, linked tail-to-head by lenses, that collectively contain the history of all schemas for a single kind of object.
 * **Instance:** Some data that is valid with respect to a schema.
-* **Lacuna:** A gap in a lens's mapping logic such that a particular instance, having passed through that lens, has some meaningfully semantic deficiency.
+* **Lacuna:** A gap in a lens's mapping logic such that a particular instance, having passed through that lens, has some meaningful semantic deficiency.
 
 These logical components are deeply interlinked as illustrated below. The first four components - schema, sequence, lens, and lineage - can be understood independent of a program that is running a useful operation. If we had a valid lineage containing two sequences, each containing two schema, and thus necessarily one explicitly-defined lens, it would look like this:
 
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -1,55 +1,53 @@
-# Thema Overview
+# About Thema
 
-Thema is a system for expressing schema for a particular kind of object and safely evolving that schema definition over time, including across what would otherwise be breaking changes. Thema consists of three kinds of things:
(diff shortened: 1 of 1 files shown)


docs/overview.md
```diff
@@ -1,80 +1,78 @@
...
-* **Sequence:** An ordered list of Schema, each being backwards compatible with its predecessors.
-* **Lens:** A bidirectional mapping between the final Schema in one Sequence, and the first Schema in another Sequence.
-* **Lineage:** An ordered list of Sequences, linked tail-to-head by Lenses, that collectively contain the history of all Schemas for a single kind of object.
-* **Instance:** Some data that is valid with respect to a Schema.
-* **Lacuna:** A gap in a Lens's mapping logic such that a particular Instance, having passed through that Lens, has some meaningfully semantic deficiency.
+* **Sequence:** An ordered list of schema, each backward compatible with its predecessors.
+* **Lens:** A bidirectional mapping between the final schema in one sequence, and the first schema in another sequence.
+* **Lineage:** An ordered list of sequences, linked tail-to-head by lenses, that collectively contain the history of all schemas for a single kind of object.
+* **Instance:** Some data that is valid with respect to a schema.
+* **Lacuna:** A gap in a lens's mapping logic such that a particular instance, having passed through that lens, has some meaningfully semantic deficiency.
```
malcolmholmes, line 24:
```suggestion
* **Lacuna:** A gap in a lens's mapping logic such that a particular instance, having passed through that lens, has some meaningful semantic deficiency.
```

docs/overview.md
```diff
@@ -1,80 +1,78 @@
...
+## About Thema Operations
 
-Thema has three essential operations, given a valid Lineage:
+Thema has three essential operations, given a valid lineage:
 
-* **Pick():** get a particular Schema from a Lineage by version number.
-* **Validate():** check whether some data is an Instance - whether it's valid with respect to a particular Schema.
-* **Translate():** given an Instance of some Schema in a Lineage, transform it to an Instance of some other Schema in the same Lineage, and emit any Lacuna arising from the transformation.
+* **Pick():** get a particular schema from a lineage by version number.
+* **Validate():** check whether some data is an instance to identify whether it's valid with respect to a particular schema.
```
malcolmholmes, line 39: Do we validate against a schema or a lineage?

sdboyer: Schema, same as in any other system.

It's a reasonable question, given the current framing. i think the other diagram will help (i have to rewrite my graffle a bit so that it really works as a GIF).

So, while data is validated to be an instance by checking against a particular schema, it's _also_ possible to [`SearchAndValidate()`](https://pkg.go.dev/github.com/grafana/thema#section-readme), where you hand some possibly-valid data to an entire lineage, which will traverse the lineage attempting to validate against all the contained schemas. 

To me, this is one of the helpers "composed from these three parts", below. But that's clearly obvious only to me, and probably deserves mention at this high level.

sdboyer: will also help with this "three step process" explanation

sdboyer (approved):
Brilliant, thank you!

Issue #28, opened 2022-01-24
Make the input kernel return validation errors based on the `To` schema


sdboyer:
Currently, when data to `InputKernel.Converge()` fails validation, it will simply return a bland `validation failed` error. This isn't helpful in any real scenario. Instead, let's have it return the validation error that came back from the `To` schema, as we can be reasonably certain that that's the schema version the author is thinking about, anyway.

i didn't do this initially because i wasn't quite sure it would fit the user's mental model, but on seeing folks actually use it, it seems clear that this is the assumption they'd make.

sdboyer:
Fixed in #34

PR #34, opened 2022-01-26
Kernel: return error for target schema in validation

kylebrandt:
for #28

### DIFF
--- a/kernel/input.go
+++ b/kernel/input.go
@@ -1,7 +1,6 @@
 package kernel
 
 import (
-	"errors"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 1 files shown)


kylebrandt:
As is, tests are not passing, need to look as to why.

sdboyer:
That's surprising, i'll have to look - it's saying that `after` isn't being assigned a value from the lens translation, and is therefore incomplete

kernel/input.go
```diff
@@ -127,11 +126,14 @@ func (k InputKernel) Converge(data []byte) (interface{}, thema.TranslationLacuna
 		return nil, nil, err
 	}
 
-	// Validate that the data constitutes an instance of at least one of the schemas in the lineage
-	inst := k.lin.ValidateAny(v)
```
sdboyer, line 131: So this call needs to be preserved - it just needs to happen after the `Validate()` call that you added.

Reasoning is that we still want to check against all possible schema inputs (which `ValidateAny()` does), we just want to try against the `To` version first, and use the error from it in the event of failure against all the schemas to improve feedback.

sdboyer (changes requested):

sdboyer:
Ugh, i need to set up CI :/

kernel/input.go
kylebrandt: Not sure I follow the logic if after the one I added, but pushed a change that that puts in before.... maybe that is what you meant?

sdboyer: it's backwards to what i intended here, but i don't think it's a big deal either way - whether it's preferable to do what you did or what i did depends on speculation (whether most users use 

A proper job would stop using `ValidateAny()` and instead construct [the loop](https://github.com/grafana/thema/blob/main/impl.go#L218-L222) ourselves, so that we never visit a schema more than once, and can choose to visit the `To` schema first. (There's a public equivalent to everything that's private in the link, so it's doable)

But I'm fine with this as written, or the way i originally described, or with creating our own loop :) Don't feel like overindexing on this rn

sdboyer (approved):

PR #37, opened 2022-02-02
thema: Introduce `AssignableTo` func and use it to fix `NewInputKernel()` type checking

### LINKED ISSUE #23: kernel: Go type incorrectly fails validation against schema
Currently, this Go type:

```go
type Example struct {
    Enum string `json:"enum"`
    FlagMap map[string]bool `json:"flagmap"`
}
```

will fail validation, as performed in `kernel.NewInputKernel()`, against this Thema schema (lineage scaffolding omitted):

```cue
{
    enum: "foo" | "bar"
    flagmap: [string]: bool
}
```

We don't want that.

We may have to work a bit at getting this really correct, but i think the basic desired check is, "will all valid instances of the schema be assignable to the Go type, and would a round-trip result in the same input." That does likely mean - as in the `enum` case above - that the Go type will allow some values that the schema does not. But, even with the impending arrival of generics in Go, i see no way around that - CUE is simply a much more precise language for expressing bounds and controls over value spaces.

sdboyer:
This change introduces a reasonably well thought-out definition of "assignability" of a Thema schema to a Go type, adding it as a func: 

```go
thema.AssignableTo(sch thema.Schema, T interface{}) error
```

The definition of assignability needs formalization, but i see no reason the Go interface type signature would change.

Fixes #23 

cc @mpvl @myitcv - this seems (?) similar to but somewhat different from what's in [gocodec](https://pkg.go.dev/cuelang.org/go@v0.4.1/encoding/gocode/gocodec). Really, this is kinda intended to be complementary to, or even a stopgap for, proper Go type codegen - "given that we hand-wrote a Go struct type, is it equivalent to what proper codegen would produce?"


kernel/input_test.go
```diff
@@ -275,3 +277,285 @@ func TestInputKernelConverge(t *testing.T) {
...
+func TestAssignable(t *testing.T) {
+	ctx := cuecontext.New()
+
+	var blah string
+	tt := map[string]struct {
+		T       interface{}
+		cue     string
+		invalid bool
+	}{
+		"nonpointer": {
```
sdboyer, line 290: @papagian @kylebrandt @radiohead if you have 15 minutes to look over this list of relations and suggest/ask about any you don't see represented, that'd be lovely. The current set could be described as "decent, but not even pretending to be exhaustive."

sdboyer:
There's plenty more to be hammered out to _really_ formalize the assignability relation here, but this gets us enough runway to work with. Leaving the rest for follow-ups.

PR #46, opened 2022-04-26
thema: Introduce [de]hydrate API

### LINKED ISSUE #43: Make [de]hydration/default trimming a part of Thema's API
Some time ago, @ying-jeanne wrote logic to trim defaults out of a Thema schema (back when Thema was still inside Grafana and called Scuemata). The basic goal there is very simple: given a valid instance of a schema, produce a new instance of the schema that has all default values {present,absent}. I copied that logic [into Thema](https://github.com/grafana/thema/blob/main/trim.go) when initially setting all this up, but kept it un-exported and haven't actually attached it to the API.

It's time to make that a proper part of the Thema API. I think it makes the most sense to attach it to the `Instance` type as a method with some rendering options, though i may change my mind once i get in there.

Some follow-up TODOs

- [ ] Add an option to de/hydrate output to `thema data translate` command
- [ ] Munge the src name of de/hydrated Instances
- [ ] Add test suite

sdboyer:
First pass at a de/hydration API.

Initially, i'd been thinking that it would work well to turn hydration into a formatting option. And that still might be worth doing. But i think this is cleanest and simplest for now - just have `Hydrate()` and `Dehydrate()` methods on `Instance` that apply the transformation and return a new instance. (Much category, very morphism!)

(cc @ying-jeanne - really this just takes her code and adapts it to this new environment! 🎉 )

Fixes #43 

---

edit: hydration is basically working, but i've futzed something up with dehydration. Going to circle back on it a little later.

Also, it may be preferable to append a suffix to the name of the input src/file (e.g. `input.json` becomes `input-hydrated.json`) when calling Hydrate() or Dehydrate(). That way future errors (if any) on that instance can be made clearer that they're happening against a modified object.


sdboyer:
Note that, as-is, this is lacking tests (similar similar to other cases) and also i'd also probably like to refactor the code a bit to make sure that it's actually not possible for these operations to error. Right now, we panic on errors, and it really should be possible to just exclude that outcome entirely, given that we already know the `Instance` is valid with respect to its `Schema`.

Gonna defer those, though.

sdboyer:
Gonna add support to the thema CLI, too

PR #50, opened 2022-05-11
encoding/jsonschema: Add JSON Schema exporter

### LINKED ISSUE #49: Add helper to export Thema schemas as JSON Schema
With #47 in, the next logical step is to also support JSON Schema as an output format. Unfortunately, there's no native helper for this in the core CUE libraries, so we'll need to get...creative.

sdboyer:
This introduces an exporter that converts a Thema schema into JSON Schema (Wright draft 4).

What this _actually_ does is run the OpenAPI exporter (which produces OpenAPI 3.0) and then does a bunch of AST hocus pocus to turn it into JSON Schema. I based this implementation on an annoyingly large amount of reading, [this JS implementation](https://github.com/openapi-contrib/openapi-schema-to-json-schema), and [these docs](https://swagger.io/docs/specification/data-models/keywords/).

There are some odds and ends to clean up:

* [x] Handle `format` oddities between OpenAPI and JSON Schema
* [x] Finish blacklisting all the oapi-specific keywords

However, with the AST manipulation pattern in place, these are pretty trivial additions.

This could possibly eventually be contributed back to CUE directly upstream. The preferable implementation there, though, is probably to update the `encoding/openapi` exporter to support oapi 3.1, which was specifically designed to be == to JSON Schema.

Fixes #49

### DIFF
--- a/encoding/jsonschema/jschema.go
+++ b/encoding/jsonschema/jschema.go
@@ -1,0 +1,433 @@
+package jsonschema
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	"cuelang.org/go/cue/ast"
+	"cuelang.org/go/cue/ast/astutil"
+	"cuelang.org/go/cue/token"
+	"github.com/grafana/thema"
+	"github.com/grafana/thema/encoding/openapi"
+)
+
+// GenerateSchema generates a JSON Schema (Draft 4) schema representation of the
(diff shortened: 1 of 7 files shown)


sdboyer:
Turns out `format` oddities aren't something we have to worry about, as the CUE stdlib openapi generator already narrows the field of possibilities to those acceptable with json schema.

PR #58, opened 2022-07-20
Ensure quoted fields when formatting to cue.Instance

sdboyer:
This is basically working around an upstream CUE issue with this:

```cue
foo: {
  bar: number
  number: string
}
```

This is valid CUE, with the LHS `number` being interpreted as a label. But currently, the OpenAPI encoder in CUE stdlib has the label shadow the type keyword. That is, it interprets the value of `bar` as a reference to the field labeled `number`, rather than as a type. (TODO: open an issue on this)

This is kind of an antipattern anyway - it's good practice to quote such a label field:

```cue
foo: {
  bar: number
  "number": string
}
```

And, when exporting to openapi (`cue def -o openapi:` with whatever params), this is fine, because cue stdlib still can actually work with instances directly. We can't, though, and we have to rely on the hack of calling `cue.Value.Syntax()` to get what we want - and the output of that method does not preserve label quotes (TODO: open an issue on this).

So, this introduces a short term hacky hack hack to just always ensure certain keyword fields are quoted.

### DIFF
--- a/encoding/openapi/oapi.go
+++ b/encoding/openapi/oapi.go
@@ -1,9 +1,6 @@
 package openapi
 
 import (
-	"fmt"
-
-	"cuelang.org/go/cue"
 	"cuelang.org/go/cue/ast"
 	"cuelang.org/go/encoding/openapi"
 	"github.com/grafana/thema"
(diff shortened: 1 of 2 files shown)


sdboyer:
The actual name for things like `number` is ["predeclared identifiers"](https://cuelang.org/docs/references/spec/#predeclared-identifiers)

which kinda explains why shadowing is a thing here

PR #63, opened 2022-08-08
Replace `kernel` package with `vmux`; introduce generics

### LINKED ISSUE #53: Make a kernel that uses generics
Thema's kernels are an ideal place to use generics: Make a kernel with type `T`; take `[]byte` in, spit `T` out.

I haven't played enough with generics yet to have a confident vision of exactly what it should look like, but i currently see no reason to believe that we shouldn't just replace the existing `InputKernel` entirely with this approach.

This is also pretty self-contained - i suspect somebody else could pick it up :)

### LINKED ISSUE #57: Change "Kernel" to "Mux"
As i was working on a larger design doc encompassing Thema, it occurred to me that a good way of describing [a main intended usage pattern for Thema's versioning system](https://github.com/grafana/thema/blob/main/docs/go-usage.md) is as a "version multiplexer". 

From [wikipedia](https://en.wikipedia.org/wiki/Multiplexer):

> A multiplexer...also known as a data selector, is a device that selects between several analog or digital input signals and forwards the selected input to a single output line.

This is a totally fair description of what Thema's kernels do. Just, rather than doing them over some set of distinct signals/"conversations"/HTTP handlers, it does it over the set of versions of schemas for a given thing. Seems like a much more descriptive name than "kernel."

sdboyer:
This introduces generics in the base thema package, allowing the pairing of a Go type with a `Schema` (`TypedSchema`) and its corresponding `Instance` (`TypedInstance`).

Building on these generics, we replace the previous `kernel` approach with a new `vmux` (version multiplexer) package. This package has the same basic goal as the original kernels, but is less work and more elegantly shaped in its final product.

Rather than having two methods (`Converge`, `ConvergeJSON`) the user calls to pass bytes through a pipeline, `vmux` contains four muxers, each of which are themselves callable functions following the same basic muxing pattern (accept all versions->see one), mapping from bytes to...

* `Untyped`: `[]byte -> *thema.Instance`
* `Byte`: `[]byte -> []byte`
* `Typed`: `[]byte -> *thema.TypedInstance[T]`
* `Value`: `[]byte -> T`

The pattern is also extensible for "middleware" in the future. The known case there is stepwise interception of `Translate` on the basis of lacuna that may have been emitted in each translation step.

Also started using some of the go1.19 godoc hotlinking conventions.

- [x] Move tests over from `kernel`
- [x] Remove `kernel` package
- [x] Clean up docs on helper funcs for generic binding
- [x] Test and implement `TypedSchema.NewT()`
- [x] Implement a lineage type corresponding to `TypedSchema`
- ~Rewrite relevant tutorials~ (deferred)
- ~Add error taxonomy for differentiating error classes~ (deferred)

Fixes #53
Fixes #57


sdboyer:
Gonna end up with `ConvergentLineage` as the typed lineage. This is gonna be a key stepping stone to what's next, which is a native code generator directly in the thema CLI.

PR #69, opened 2022-09-12
Update cue `crd.spec.versions` to an array

IfSentient:
Convert `crd.spec.versions` to an array of objects to conform with CRD definition, and avoid errors with multiple versions.


sdboyer:
derp - thanks!

PR #75, opened 2022-10-20
Quickstart doc for using Thema in Go

### LINKED ISSUE #74: Rewrite initial Thema tutorial to focus on just basic schema authoring, then pivot to codegen right away
Thema's existing tutorial (parts [1](https://github.com/grafana/thema/blob/main/docs/authoring.md), [2](https://github.com/grafana/thema/blob/main/docs/go-mapping.md), and [3](https://github.com/grafana/thema/blob/main/docs/go-usage.md)) is outdated and overly verbose.

All of part 2 and most of part 3 are now now pretty much done by codegen (`thema lineage gen`). `thema lineage gen gobindings` with `--bindtype` even goes a step beyond part 3, thanks to generics and the new vmux system.

I think a new pass at the tutorials should focus on getting a basic thema-based dev loop going: write a schema, update types and bindings in Go and TS, test with some data.

Subcommands within `thema lineage init`, `thema lineage gen`, `thema data validate` should be sufficient for all this. We can, for now, ignore multiple schemas and lenses.

Some of what's in the existing tutorials could probably be reused, but if we just started from scratch, i wouldn't be unhappy 😄

ishanjainn:
Fixes #74


docs/go-quickstart.md
```diff
@@ -0,0 +1,168 @@
...
+import "github.com/grafana/thema"
+
+thema.#Lineage
+name: "ship"
+seqs: [
+	{
+		schemas: [
+			// v0.0
+			{
+				firstField: string
```
sdboyer, line 47: Let's definitely include some field comments right from the start, so that the user immediately has the impression that documenting their fields is important.

sdboyer (changes requested):
Looks like a great start! 🎉 🎉 

Gonna do another pass in a bit, but what if, in the ship, we actually try to describe a [ship](https://en.wikipedia.org/wiki/Full-rigged_ship)?

So, instead of:

```cue
firstField: string
secondField: int // or int32, per note about precision rules
```

We do something like:

```cue
// name is what we call the ship, and what's written in big letters on its hull
name: string
// masts is the number of masts the ship has. No fully rigged ship
// has ever had more than 7: https://oceannavigator.com/the-most-masted-schooner-ever-built/
masts: uint8 & < 8
```

This provides some generated comments, as well as a type that Go's type system can't fully capture.

docs/go-quickstart.md
```diff
@@ -0,0 +1,168 @@
...
+package ship
+
+// Ship defines model for ship.
+type Ship struct {
+	FirstField  string `json:"firstField"`
+	SecondField int    `json:"secondField"`
+}
+```
+
+If we add another schema to the sequence in `ship.cue` and again generate the Go type, Go will lose track of the first schema and will represent the second schema only.
```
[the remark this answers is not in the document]
ishanjainn: Got it

docs/go-quickstart.md
```diff
@@ -0,0 +1,167 @@
...
+	//     for example, declared schemas don't follow backwards compatibility rules,
+	//     lenses are incomplete.
+	return thema.BindLineage(raw, rt)
+}
+
+// type guards
+
+var _ thema.LineageFactory = Lineage
+```
+This sets us up where we have a single Go function, Which we can call and it will load up the `ship.cue` from disk and return a thema Lineage
```
sdboyer, line 149:
```suggestion
We now have a single Go function our program can call, and it will load up the `ship.cue` from disk and return a [`thema.Lineage`](https://pkg.go.dev/github.com/grafana/thema#Lineage)
```

docs/go-quickstart.md
```diff
@@ -0,0 +1,167 @@
...
+## Validating data
+
+We can now define a function which takes the Lineage returned from the `Lineage` function, Select the schema which we want to validate our data against, and then validate the data
+
+```go
+func validateInput() {
+	lin, _ := Lineage(thema.NewRuntime(cuecontext.New()))
+	sch, _ := lin.Schema(thema.SV(0,0))  //thema.SV(0,0) here represents first schema of first sequence
+	sch.Validate(...)
+}
```
sdboyer, line 161: i think this section could be stronger if it talked about validation from both the CLI and Go, to illustrate the symmetry between the approaches.

Using the different schema fields i suggested in the other comment, you could add a file to this PR containing the following JSON:

```json
{
    "name": "Theseus",
    "masts": 9,
}
```

Then ask the reader to 
```bash
curl $URL > invalid.json
thema data validate -l ship.cue invalid.json
```

Then, suggest writing a go test file that does the above here in an actual go test. That way they can easily execute and see the results. A [byte slice embed](https://blog.jetbrains.com/go/2021/06/09/how-to-use-go-embed-in-go-1-16/) could save a few lines in the test file by reusing the downloaded file.

sdboyer: Also, an easy way to get the desired `cue.Value` is to use a [JSON Endec](https://pkg.go.dev/github.com/grafana/thema/vmux#NewJSONEndec)

docs/go-quickstart.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+We can pass a value of type `cue.Value` to `sch.Validate()` to validate the data against the first schema of the first sequence. You can also use the method `ValidateAny()` to validate the input data against any schema in our Lineage.
+
+## Wrap up
+This tutorial demostarted how you can create an empty Lineage using the thema CLI, define a simple schema and use it to generate Go types and bindings, and use the bindings to validate data against a selected schema in the Lineage.
```
sdboyer, line 167:
```suggestion
This tutorial demonstrated how you can create an empty Lineage using the thema CLI, define a simple schema and use it to generate Go types and bindings, and use the bindings to validate data against a selected schema in the Lineage. 
```

sdboyer (changes requested):
OK, i've now looked at the whole file 😄 

i hope the bunch of comments aren't sending the wrong signal. - i think this PR fundamentally _right_! I'm delighted at how many fewer words you used than i did in the tutorials i wrote. The only reason i'm even able to make suggestions is because you've framed it well.

docs/go-quickstart.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Generating Go Bindings
+
+Use the below command to generate a Go binding from `ship.cue`
+
+```
+thema lineage gen gobindings -l ship.cue > go_bindings.go
+```
+
+The genearted Go binding will look like below
```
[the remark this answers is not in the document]
ishanjainn: Yup agreed!

docs/go-quickstart.md
ishanjainn: Oo yup, will add this, definitely gives a better understanding

docs/go-quickstart.md
```diff
@@ -0,0 +1,168 @@
...
+			// v0.0
+			{
+				// TODO (delete me - first schema goes here!)
+			},
+		]
+	},
+]
+```
+
+Let's define a simple schema as an object containing two fields named `firstField` of type `string` and `secondField` of type `int`,
```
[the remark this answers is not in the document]
ishanjainn: Yup, Used the ship example above

docs/go-quickstart.md
ishanjainn: Yup added comments as you mentioned in the ship example

sdboyer:
The errors this produces are a result of #77. Rather than waiting for a fix there, i think the simplest solution for this PR is to just add instructions for the user to run `cue mod init`.

docs/go-quickstart.md
```diff
@@ -0,0 +1,184 @@
...
+```
+We now have a single Go function our program can call, and it will load up the `ship.cue` from disk and return a [`thema.Lineage`](https://pkg.go.dev/github.com/grafana/thema#Lineage)
+We can then use the Lineage function to validate our data against the schema of choice
+
+## Validating data
+
+Using the Thema CLI, Users can validate the schema of choice defined in `ship.cue`. Use the below command to validate the schema against an invalid JSON data 
+
+```
+curl $URL > invalid.json
```
malcolmholmes, line 137: What is $URL?

malcolmholmes (approved):
This tutorial was very helpful to read. It made a number of complicated things simple. Thank you for writing it.

docs/go-quickstart.md
ishanjainn: Placeholder for now, but added the Url for once the PR is merged

sdboyer (approved):

sdboyer:
@ishanjainn not sure what the issue with the CLA is - could you make sure you've signed with the relevant email? or rebase the commits in this PR to only use an account that's signed the CLA?

ishanjainn:
Closed this in favour of https://github.com/grafana/thema/pull/78

PR #82, opened 2022-10-31
thema: Flatten schemas to one-dimensional array

### LINKED ISSUE #2: Constrain sequence lenses so that an absent declaration is not valid
A key invariant of scuemata is that authors must define lenses for all sequences after the first. While it's impossible to generically verify the completeness of the translations in these lenses, or the lacunae the may emit, the bare minimum here is erroring if the author doesn't define a lens, at all.

Unfortunately, when we take the `#Lens` definition

https://github.com/grafana/thema/blob/b7b433093f2fd56c11e68a45b0c3ee7b7850ca8c/lineage.cue#L48-L67

and reference it directly in `seqs` where we want lens instances to appear:

https://github.com/grafana/thema/blob/b7b433093f2fd56c11e68a45b0c3ee7b7850ca8c/lineage.cue#L72-L80

it means that a lineage is still valid even when the author omits the field. The questions are:

* What's a reasonable thing to force a constraint failure on? Should `#Lens` itself be refactored so that it's more amenable to doing this?
* How do we square wanting this to be incomplete with _also_ wanting to provide helpers that reduce some of the boilerplate of writing lenses?

Maybe this is just yet another place where we need a `must()` check (cue-lang/cue#943), particularly to emit the right error.

Note that this issue is targeting a relatively minimal check - to climb up to our true-goal invariant, in addition to basic lens existence, we have to be certain that a) the Lens connects the end of the prior sequence to the head of its current sequence, and b) that given a concrete input valid wrt the predecessor schema, the `rel` of both forward and reverse lenses is sufficient to produce a concrete output of the successor schema.

### LINKED ISSUE #18: Refactor `#Lineage` to support reverse lens-mappings between schema within a sequence
Currently, lenses exist solely at the `#Sequence` level, and only for sequences after the first:

https://github.com/grafana/thema/blob/b7b433093f2fd56c11e68a45b0c3ee7b7850ca8c/lineage.cue#L69-L80

This is a relic of older thinking, where i was trying to convince myself that it's sufficient to guarantee translatability to the _latest_ schema. It's not - if the same guarantee doesn't hold in the reverse direction, then a familiar evolutionary coupling results for communicating systems: it's not safe to start sending a new schema version until all receivers are known to have updated to accept it. It's arguably not as bad as the primary case, where receivers can't update until senders are known to support it, as receivers are often the hubs - but it's still an unacceptable degree of coupling.

Backward compatibility does not imply forwards compatibility/is a non-invertible relation, so introducing a requirement for reverse translatability means we have to change how lineages and sequences themselves are declared in order to introduce a place for the in-sequence reverse lenses.

Fixing this requires changing something fundamental about how lineages are declared. The simplest answer would be introducing an additional reverse-lens list onto `#Sequence` that's `len(schemas)-1`, though i haven't actually tried that to see what other weird warts that produces.

Even if that does work, i'm not sure i'm happy with the ergonomics - lineages are already a deeply intermeshed datastructure, and it's hard enough to keep track of everything going on in them. It's possible that we should use this additional requirement to trigger revisiting the way lineages are structured (in a way that preserves key invariants, obviously). I need to put up a discussion around that.

---

**Although...**

At least...i _think_ we have to do this, but i'm not actually sure. I have this nagging voice in the back of my mind, saying that the set of changes allowable under backwards compatibility/subsumption rules actually _do_ imply a lens morphism that can be inferred, at least in almost all cases:

* If a new field was added, drop the field in the translated instance
    * This is the approach irrespective of whether the added field is optional or required+default
* If a field contains a disjunction with a concrete default (e.g. `field: *"foo" | "bar:`), and new branches were added to the disjunction (`field: *"foo" | "bar" | "baz"`), backwards compatibility rules already dictate that the default (`*"foo"`) must exist in prior versions. If an instance containing a value from a newly-added branch (`"baz"`) is reverse-translated, fall back to the default (`"foo"`)
  * If a field contains a disjunction with a non-concrete default or no default at all, and the newer instance falls outside the acceptable value space for the older schema, we cannot reach a resolution.

If the above is correct, it's not complete _enough_ to obviate the need for a changed structure - though it could mean that automated logic is applied in the absence of a declaration. I'm loathe to get too far into lens generation right away, as it's probably a big rabbit hole, but i'd be remiss to not at least mention the possibility.

sdboyer:
After a suggestion from @myitcv at Kubecon last week, the light finally clicked on with how we can get rid of the nested array in `seqs`, add reverse lenses to the structure, and overall just drastically improve UX. In fact, we can get rid of the concept of "sequence" entirely.

As an example, the final form for the [narrowing exemplar](https://github.com/grafana/thema/blob/main/exemplars/narrowing.cue) is this:

```cue
schemas: [{
	version: [0, 0]
	schema: boolish: "true" | "false" | bool | string
}, {
	version: [1, 0]
	schema: properbool: bool
}]

lenses: [{
	to: [0, 0]
	from: [1, 0]
	input: _
	result: {
		// Preserving precise original form is a non-goal of thema in general.
		boolish: input.properbool
	}
	lacunas: []
}, {
	to: [1, 0]
	from: [0, 0]
	input: _
	result: {
		if ((input.boolish & string) != _|_) {
			properbool: input.boolish == "true"
		}
		if ((input.boolish & bool) != _|_) {
			properbool: input.boolish
		}
	}
	lacunas: [
		{
			sourceFields: [{
				path:  "boolish"
				value: input.boolish
			}]
			targetFields: [{
				path:  "properbool"
				value: result.properbool
			}]
			message: "boolish was a string but neither \"true\" nor \"false\"; fallback to treating as false"
			type:    thema.#LacunaTypes.LossyFieldMapping
			cond:    ((input.boolish & string) != _|_) && ((input.boolish & ("true" | "false")) == _|_)
		},
	]
}]
```

`seqs` is gone, nested arrays are gone. `version` is an explicit field within the `#SchemaDef` (though there's still no actual choice to be made by the user - it's all governed by backwards compat rules) Additional fields are possible in there, including providing examples of the schema.

Lenses are moved into their own, separate structure, independent of the `schemas` list. They're a bit more verbose than i'd like, with it still be necessary to explicitly write the `input` field. But `from` and `to` are now just a version number, rather than needing to be a reference to the correct schema. Better names for the other fields, too - i think `result` is indicative of what's supposed to go there, though i don't think this has reached the level of self-explanatory quite yet.

Fixes #2 
Fixes #18

- [ ] Add native CUE check that lens set is complete
- [x] Add more native CUE test cases 
- [x] Convert existing exemplars, tests
- [x] Refactor Go packages to work with new CUE structures
- [x] Introduce AST-based migrator


lineage.go
```diff
@@ -5,173 +5,100 @@ import (
...
 var (
-	_ Lineage                     = &UnaryLineage{}
+	_ Lineage                     = &baseLineage{}
 	_ ConvergentLineage[Assignee] = &unaryConvLineage[Assignee]{}
 )
 
-// A UnaryLineage is a Go facade over a valid CUE lineage that does not compose
+// A baseLineage is a Go facade over a valid CUE lineage that does not compose
 // other lineage.
-type UnaryLineage struct {
```
spinillos, line 19: It was used in [Kindsys](https://github.com/grafana/kindsys/blob/812b9884a375b6cdf20aa9c3f9676dd2b770df8b/somedef.go#L28) library. We need to update that too.

sdboyer:
This is nearly done. We're going to defer a bunch of work to follow-ups, including:

* Create a consistent error model using `github.com/cockroachdb/errors`
* Add support for including examples in `#SchemaDef`
* Rewrite and re-export most of the `encoding/cue` package, which was largely pulled out here
* Tweak base CUE definitions to improve evaluator performance
* Remove old txtar testing library

sdboyer:
This is finally, _finally_ good to go. Lots of follow-ups, of course.

PR #84, opened 2022-11-01
cli: Generate files adjacent to input, with sane cuefs

### LINKED ISSUE #77: Output of `thema lineage gen gobindings` does not work outside a `cue.mod`
The CLI go bindings generator (`thema lineage gen gobindings`) produces a `cueFS` that tries to embed the right files. Even when it does, though, there's a problem - [`load.InstancesWithThema()` requires a](https://pkg.go.dev/github.com/grafana/thema@v0.0.0-20221019174755-33088a06e0ad/load#InstancesWithThema) `cue.mod/module.cue` to exist at the root of the `modFS` that's passed in.

Or at least, that's the idea. But because of how CUE's `load.Instances()` scans the filesystem, `load.InstancesWithThema()` _usually_ works even without that file so long as it's being executed with a cwd that either has a `cue.mod`, or one of its parent dirs does. This creates an awkward state where `thema lineage gen gobindings` creates working code _sometimes_ - on the developer's local, when there's a `cue.mod`.

We need a more robust approach to generation that can work with defaults for the simple/demo case, but still be useful when the codebase in which it's being used has a larger `embed.FS` system in which the generated code exists. Grafana has [such a system](https://github.com/grafana/grafana/blob/main/embed.go#L10), and has built [its own helpers](https://github.com/grafana/grafana/blob/main/pkg/cuectx/ctx.go#L63) - though they aren't great either, and this all really needs genericizing.

### LINKED ISSUE #81: cli: Generate Go types and bindings in an adjacent file
Currently, the Go types and bindings generator in the CLI generate to stdout. This isn't terribly idiomatic - proto sets the general pattern here, with generating an adjacent file. It's probably generally more convenient for callers, anyway.

This is especially problematic for bindings, though, because the generating an `embed.FS` that will reliably work when passed to `load.InstancesWithThema` (or any such function we could write) requires being able to inspect the disk environment where the generated file will land. That's impossible on stdout.

So we'll have both of those generators default to generating an adjacent file (`<name>_types_gen.go`, `<name>_bindings_gen.go`), and if bindings are forced to stdout, just not allow an embed to be generated at all - it's up to the caller to make it sane.

sdboyer:
Numerous fixes to the go code generator for both generating output adjacent to the input lineage file, and generating a sane cueFS that'll actually WORK given cue.mod environs

(still finishing the latter part)

Fixes #81 
Fixes #77


sdboyer:
Got the FS generation sorted out, finally. It involved introducing some function call indirection, which is toggleable by config inputs. But i think we're in a decent enough place, now.

Issue #54, opened 2022-06-21
Use "major version" and "minor version" for syntactic versions


sdboyer:
Thema's versioning system has two digits, e.g. `v0.0` - what people coming from semver would be naturally inclined to call a "major version" and a "minor version."

i resisted this naming initially, for reasons i can't even remember. Instead, i've been calling the first number "sequence version" and the second "schema version." But this was an error. For one, "schema version" is confusingly ambiguous - is that the second number, or the version number as a whole? And second, the cost of asking people to learn new terms for version number positions is significantly higher than the risk of having folks transfer possibly-slightly-incorrect associations from semver.

Putting me over the top on this was the fact that [OpenMetadata](https://docs.open-metadata.org/metadata-versioning/metadata-versioning) landed on `major.minor` with exactly the same semantics.

So - all docs need to be updated to use `major` and `minor` where appropriate. Variables and fields should be renamed from e.g. `seqv` and `schv` to `majv` and `minv`.

frossano-grafana:
@rhorvath  can we look at this PR and see if it's good to be merged, and close the issue?

PR #87, opened 2022-11-03
Use "major version" and "minor version" for syntactic versions

ishanjainn:
This PR makes to use `major` and `minor` version instead of Sequence and Schema versions naming where appropriate. Variables and fields are renamed from `seqv` and `schv` to `majv` and `minv`

Resolves https://github.com/grafana/thema/issues/54


ishanjainn:
@sdboyer 
I have updated the docs but wanted to take your opinion if the docs part makes the version part clearer?

docs/go-usage.md
```diff
@@ -121,7 +121,7 @@ func TestHandpickValidation(t *testing.T) {
 }
 ```
 
-Here, we've hand-picked the schema version we want to validate against - `0.0`, which every lineage is guaranteed to contain. The [`LatestVersion()`](https://pkg.go.dev/github.com/grafana/thema#LatestVersion) and [`LatestVersionInSequence()`](https://pkg.go.dev/github.com/grafana/thema#LatestVersionInSequence) functions provide fuzzier version selection logic. But allowing only one schema version as input somewhat defeats the purpose of using Thema in the first place. Ideally, we'd have something more dynamic.
+Here, we've hand-picked the syntactic version of the schema we want to validate against - `0.0`, which every lineage is guaranteed to contain. The [`LatestVersion()`](https://pkg.go.dev/github.com/grafana/thema#LatestVersion) and [`LatestVersionInSequence()`](https://pkg.go.dev/github.com/grafana/thema#LatestVersionInSequence) functions provide fuzzier version selection logic. But allowing only one syntactic version of a schema from the Lineage of theas input somewhat defeats the purpose of using Thema in the first place. Ideally, we'd have something more dynamic.
```
sdboyer, line 124:
```suggestion
Here, we've hand-picked the syntactic version of the schema we want to validate against - `0.0`, which every lineage is guaranteed to contain. The [`LatestVersion()`](https://pkg.go.dev/github.com/grafana/thema#LatestVersion) and [`LatestVersionInSequence()`](https://pkg.go.dev/github.com/grafana/thema#LatestVersionInSequence) functions provide fuzzier version selection logic. But allowing only one syntactic version as input somewhat defeats the purpose of using Thema in the first place. Ideally, we'd have something more dynamic.
```

sdboyer (commented):
A number of these changes make explicit that when the docs are talking about version numbers, it's a syntactic version, of a schema, in a lineage.

I'm undecided about this. When i wrote the docs originally, i thought, "it's preferable to assume that the reader has understood that all versions in thema are syntactic versions, and thus write shorter, more readable sentences that leave that implicit."

However, i also recognized that i'm a terrible judge of whether that's a reasonable assumption or not, because i have the [curse of knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge) here. So i'm happy to defer - do you find it preferable to be more verbose, explicitly stating that versions are syntactic versions?

ishanjainn:
Wasnt in the favour of being more verbose, Instead Ill prefer just explain as a note within the first doc the user reads(overview.md) what `schema version` means and continue using it thereafter. Let me

PR #90, opened 2022-12-15
Improve OpenAPI generator

sdboyer:
Refactor codegen to no longer rely on eval and printing, taking
advantage of the ability to pass a cue.Value to the CUE standard library
package.

Also introduce support for a group flag, which allows translating only the
children of the root schema to OpenAPI.

And introduce a long-overdue test txtar-based test framework.

cc @IfSentient


sdboyer:
cc @undef1nd - fyi, this testing system should be equally useful for what you've been looking at with JSON Schema generation

sdboyer:
This testing system is 🔥 . It can treat the exemplar set as a corpus for tests using only a single flag. Add a new exemplar to the corpus, and it instantly flows to all tests using the framework.

sdboyer:
I've also added a `cfg.Subpath`, which allows forcing the generator to gen just one subpath within the provided schema, rather than the whole thing.

This is mostly ready to go. The nasty bug with extra files being appended to txtar when golden updates are on remains, and i also have to figure out why some Go types aren't rendering in grok.

Once those are licked, though, this is ready to merge.

sdboyer:
This is in good shape. But there are some follow-ups.

I had to fork CUE for some fixes - there were a couple key cases that weren't covered (conjuncts in numeric value types, selectors on object types). This substantially impacted Go codegen elsewhere, being the difference between obviously incorrect and probably correct.

The test framework itself is also in a messy, halfway state. In retrospect, i realize what i was really trying to do was graft a subtest layer on top of the txtar-specific test runner. The framework itself [already creates a subtest per txtar](https://github.com/grafana/thema/blob/392072f53f26286e560966967b00b5566fb59a23/internal/txtartest/txtar.go#L454), but i was effectively trying to make each provided subtest executed from there be able to construct its own subtests in a way that's basically similar to stdlib `*testing.T.Run`.

Given that the central object in the framework is the lineage, this is crucial to allow for [more standard helpers](https://github.com/grafana/thema/blob/392072f53f26286e560966967b00b5566fb59a23/internal/txtartest/helpers.go#L19) that, for example, allow [a particular test case](https://github.com/grafana/thema/blob/392072f53f26286e560966967b00b5566fb59a23/encoding/openapi/oapi_test.go#L70) to be executed across all schemas in a lineage, and keeping each subtest's output therein tidily namespaced.

PR #95, opened 2022-12-22
Update cuetsy to 0.1.3

spinillos:
It fixes maps 🤞🏻 in TS


spinillos:
I'm going to close it because it breaks models like 
```cue
{
    foo: "bar"
    ...
}
```
generating `unknown` instead of `{ foo: 'bar'; }`

PR #98, opened 2023-01-13
Replace ast with dst for gocode

spinillos:
It updates ast with dst for go files.


encoding/gocode/gen.go
```diff
@@ -42,10 +42,10 @@ type TypeConfigOpenAPI struct {
 	// lowercase version of the Lineage.Name() is used.
 	PackageName string
 
-	// ApplyFuncs is a slice of AST manipulation funcs that will be executed against
+	// ApplyFuncs is a slice of DST manipulation funcs that will be executed against
```
sdboyer, line 45:
```suggestion
	// ApplyFuncs is a slice of AST manipulation funcs that will be executed against
```

i think best to leave this one as "AST", as this is referring to the general concept of abstract syntax trees, not the particular implementation.

encoding/gocode/gen.go
```diff
@@ -218,10 +218,10 @@ type BindingConfig struct {
...
```
sdboyer, line 221:
```suggestion
	// ApplyFuncs is a slice of AST manipulation funcs that will be executed against
```

same as above

encoding/gocode/gen.go
```diff
@@ -397,7 +397,7 @@ func postprocessGoFile(cfg genGoFile) ([]byte, error) {
 
 		if len(added) != 0 {
 			// TODO improve the guidance in this error if/when we better abstract over imports to generate
-			return nil, fmt.Errorf("goimports added the following import statements to %s: \n\t%s\nRelying on goimports to find imports significantly slows down code generation. Either add these imports with an AST manipulation in cfg.ApplyFuncs, or set cfg.IgnoreDiscoveredImports to true", cfg.path, strings.Join(added, "\n\t"))
+			return nil, fmt.Errorf("goimports added the following import statements to %s: \n\t%s\nRelying on goimports to find imports significantly slows down code generation. Either add these imports with an DST manipulation in cfg.ApplyFuncs, or set cfg.IgnoreDiscoveredImports to true", cfg.path, strings.Join(added, "\n\t"))
```
sdboyer, line 400:
```suggestion
			return nil, fmt.Errorf("goimports added the following import statements to %s: \n\t%s\nRelying on goimports to find imports significantly slows down code generation. Either add these imports with an AST manipulation in cfg.ApplyFuncs, or set cfg.IgnoreDiscoveredImports to true", cfg.path, strings.Join(added, "\n\t"))
```

Same other comments

sdboyer (changes requested):
Just small changes in comments

sdboyer:
[This line](https://github.com/grafana/thema/pull/98/files#diff-cadaec9c1293e707e5ddcfd1461963d19ac4cdbfad67116a957f4e7782d91874R374) made me crave some tests that even just minimally verify that output hasn't changed. i realize there are no tests for this right now - totally my fault - but maybe this is a good opportunity to add some.

sdboyer:
Basically a txtar-based test like what we have for [openapi](https://github.com/grafana/thema/blob/main/encoding/openapi/oapi_test.go) should do the trick - something that shows outputs of the generator under different parameterizations of inputs

sdboyer (approved):
LGTM! thanks for adding the tests

PR #102, opened 2023-01-18
Depointerizer types

spinillos:
It moves `DecoderCompactor` from Grafana and create `Depointerizer` function to remove pointers from the types.

By default, we are going to remove always the pointers in maps and slices but in the rest of the types will be optional. So the user could decide to remove these pointers or not setting `NoOptionalPointers` in the configuration.

Related: https://github.com/grafana/grok/issues/1

### DIFF
--- a/encoding/gocode/decorators.go
+++ b/encoding/gocode/decorators.go
@@ -1,0 +1,112 @@
+package gocode
+
+import (
+	"go/token"
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/dstutil"
+)
+
+func isSingleTypeDecl(gd *dst.GenDecl) bool {
+	if gd.Tok == token.TYPE && len(gd.Specs) == 1 {
+		_, is := gd.Specs[0].(*dst.TypeSpec)
+		return is
+	}
+	return false
+}
+
+func isAdditionalPropertiesStruct(tspec *dst.TypeSpec) (dst.Expr, bool) {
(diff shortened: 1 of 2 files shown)


encoding/gocode/decorators.go
```diff
@@ -0,0 +1,112 @@
...
+
+func isAdditionalPropertiesStruct(tspec *dst.TypeSpec) (dst.Expr, bool) {
+	strct, is := tspec.Type.(*dst.StructType)
+	if is && len(strct.Fields.List) == 1 && strct.Fields.List[0].Names[0].Name == "AdditionalProperties" {
+		return strct.Fields.List[0].Type, true
+	}
+	return nil, false
+}
+
+func DecoderCompactor() dstutil.ApplyFunc {
```
sdboyer, line 26: Let's unexport this, i don't think there's value in exposing it

sdboyer (changes requested):
Let's also add some test cases that exercise the depointerizing config options.

encoding/gocode/decorators.go
spinillos: That's true, thnx

sdboyer (approved):

sdboyer:
I'll defer to a follow-up adding test cases to cover these, as we're quite short on time right now and it's not atypical of other things in thema (unfortunately 😭) that they're not tested

PR #104, opened 2023-01-25
Fix names with spaces

spinillos:
Fixes https://github.com/grafana/grafana/pull/62018/files#r1086601395

Sometimes the root name contain spaces and it generates wrong interface names in typescript.


spinillos:
We are going to add this fix inside Grafana instead: https://github.com/grafana/grafana/pull/62144

Issue #103, opened 2023-01-22
Let's also add some test cases that exercise the depointerizing config options.


sdboyer:
> Let's also add some test cases that exercise the depointerizing config options.

_Originally posted by @sdboyer in https://github.com/grafana/thema/pull/102#pullrequestreview-1263771003_

PR #105, opened 2023-01-30
Add tests for depointerizer and fixes

spinillos:
Contributes: https://github.com/grafana/thema/issues/103

It adds txtar tests for `depointerizer` and fixes. I set a map of `map[dst.Expr]bool` and `depointers[s]` was always false 🤡.
Because we have only the use cases to depointerize all or only Map/Array types, I change it with a simple bool.

Idk why map is completely ignore in group [here](https://github.com/grafana/thema/compare/tests-for-depointerizer?expand=1#diff-732bbcd269f0aa495695654a273fa240831261310998f92009d20ec26e4e2741R22-R35).

### DIFF
--- a/encoding/gocode/testdata/pointers.txtar
+++ b/encoding/gocode/testdata/pointers.txtar
@@ -1,0 +1,57 @@
+-- in.cue --
+package generate
+
+import "github.com/grafana/thema"
+
+thema.#Lineage
+name: "pointers"
+seqs: [
+        {
+            schemas: [
+                {
+				    astring?: string
+					anint?:   int
+					abool?:   bool
+					alist?: [...string]
+					amap?: [string]: string
+				},
+			]
+		},
+	]
+
+-- out/generate/0.0/group.go --
+package defaultchange
+
+// Abool defines model for abool.
+type Abool bool
+
+// Alist defines model for alist.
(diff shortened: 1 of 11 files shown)


encoding/gocode/testdata/pointers.txtar
```diff
@@ -0,0 +1,57 @@
...
+name: "pointers"
+seqs: [
+        {
+            schemas: [
+                {
+				    astring?: string
+					anint?:   int
+					abool?:   bool
+					alist?: [...string]
+					amap?: [string]: string
```
sdboyer, line 16: let's:

* Separate this single case out into its own test
* Manually write the expected output (so the test will fail when run)
* Disable the new test using [the `ToDo` map](https://github.com/grafana/thema/blob/e665e4967517c5a2bbb7ac2f2f38d9a5f70441b1/internal/txtartest/txtar.go#L68)

sdboyer (changes requested):
Thanks for the tests! Let's separate the one out per comment then open a follow-up issue

Big question is whether the bug is in cue stdlib openapi gen, or oapi-codegen, or somewhere in between with us

sdboyer (approved):
great, thank you!

PR #107, opened 2023-02-06
Upgrading oapi library

spinillos:
_**Important**_: The library introduced a bug checking pointers equality and some cue files fail. I created a PR in their repo to fix that. If they don't respond in time, we can use my fork to make it work.

It upgrades oapi library to `v1.12.4`. I'm going to try to clarify the changes...

The oapi library introduce these changes:
* **Adds GoDecl in `some` comments**: It means that it set the name of each field/struct at the beginning of each comment. That's "correct" because it makes GoLint happy but it breaks most of our comments. It works different when is group or not 🤷🏻‍♀️. 
* **Generic fields are now structs with functions to map them**: It adds a lot of boilerplate to map the struct into their different types and it adds new imports. Previously we have `*interface{}` that it did the work.
* **Fixes map structs**: We don't need to iterate the Go file to detect the maps with `"AdditionalProperties"` to fix them.
* **"Generic" functions and autogenerated structs use underscores**: We started to have structs names with underscores 🙃.  Some of our `*interface{}` became structs that its ok.

Changes that I added:
* `decoderCompactor()` isn't necessary anymore.
* It adds a `UseGoDeclInComments` to allow default configuration in comments. By default it leaves the comments "as" we put in the cue file. Its not perfect but its accurate 😞.
* It iterates the Go file to detect the structs that previously were `*interface{}` and leave them as this. Also deletes the related functions and models affected.

Pending stuff:
* Autogenerated structs that aren't generic uses underscores and we should fix them. If this PR its ok to merge, I'll create an issue to fix this case.

### "Generic" structs
When I was talking about generic structs I mean the ones that accept multiple values like:
```cue
// TODO docs
refresh?: string | false @grafanamaturity(NeedsExpertReview)
```

It was generating the following:

```go
// DashboardRefresh0 defines model for .
type DashboardRefresh0 = interface{}

// DashboardRefresh1 defines model for .
type DashboardRefresh1 = string

// Dashboard_Refresh TODO docs
type Dashboard_Refresh struct {
	union json.RawMessage
}

// AsDashboardRefresh0 returns the union data inside the Dashboard_Refresh as a DashboardRefresh0
func (t Dashboard_Refresh) AsDashboardRefresh0() (DashboardRefresh0, error) {
	var body DashboardRefresh0
	err := json.Unmarshal(t.union, &body)
	return body, err
}

// FromDashboardRefresh0 overwrites any union data inside the Dashboard_Refresh as the provided DashboardRefresh0
func (t *Dashboard_Refresh) FromDashboardRefresh0(v DashboardRefresh0) error {
	b, err := json.Marshal(v)
	t.union = b
	return err
}

// MergeDashboardRefresh0 performs a merge with any union data inside the Dashboard_Refresh, using the provided DashboardRefresh0
func (t *Dashboard_Refresh) MergeDashboardRefresh0(v DashboardRefresh0) error {
	b, err := json.Marshal(v)
	if err != nil {
		return err
	}

	merged, err := runtime.JsonMerge(b, t.union)
	t.union = merged
	return err
}

// AsDashboardRefresh1 returns the union data inside the Dashboard_Refresh as a DashboardRefresh1
func (t Dashboard_Refresh) AsDashboardRefresh1() (DashboardRefresh1, error) {
	var body DashboardRefresh1
	err := json.Unmarshal(t.union, &body)
	return body, err
}

// FromDashboardRefresh1 overwrites any union data inside the Dashboard_Refresh as the provided DashboardRefresh1
func (t *Dashboard_Refresh) FromDashboardRefresh1(v DashboardRefresh1) error {
	b, err := json.Marshal(v)
	t.union = b
	return err
}

// MergeDashboardRefresh1 performs a merge with any union data inside the Dashboard_Refresh, using the provided DashboardRefresh1
func (t *Dashboard_Refresh) MergeDashboardRefresh1(v DashboardRefresh1) error {
	b, err := json.Marshal(v)
	if err != nil {
		return err
	}

	merged, err := runtime.JsonMerge(b, t.union)
	t.union = merged
	return err
}

func (t Dashboard_Refresh) MarshalJSON() ([]byte, error) {
	b, err := t.union.MarshalJSON()
	return b, err
}

func (t *Dashboard_Refresh) UnmarshalJSON(b []byte) error {
	err := t.union.UnmarshalJSON(b)
	return err
}
```

and used like:
```go
// TODO docs
Refresh *Dashboard_Refresh `json:"refresh,omitempty"`
```

And after all changes, we are going to have the previous behaviour:
```go
// TODO docs
Refresh *interface{} `json:"refresh,omitempty"`
```


sdboyer:
cc @IfSentient

sdboyer (approved):
This looks good! Thanks for all your work on this, i know it's been quite a journey 😅

PR #109, opened 2023-02-07
More fixes for oapi codegen

spinillos:
https://github.com/grafana/thema/pull/107 misses the maps and their iterations 😥. Lucky that we have enough cue files in Grafana to test 😄

### DIFF
--- a/encoding/gocode/testdata/generics.txtar
+++ b/encoding/gocode/testdata/generics.txtar
@@ -1,0 +1,97 @@
+-- in.cue --
+package generate
+
+import "github.com/grafana/thema"
+
+thema.#Lineage
+name: "defaultchange"
+seqs: [
+        {
+            schemas: [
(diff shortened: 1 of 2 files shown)


encoding/gocode/testdata/generics.txtar
```diff
@@ -0,0 +1,112 @@
...
+
+import "github.com/grafana/thema"
+
+thema.#Lineage
+name: "defaultchange"
+seqs: [
+        {
+            schemas: [
+                {
+                    #Generic: string | bool
```
sdboyer, line 12: I think we should avoid using the word "Generic" here (and in general) in order to avoid anyb possible ambiguity with language generics

sdboyer (changes requested):

sdboyer (approved):

PR #111, opened 2023-02-10
Do a `checkscalar` on NumberKind instead of erroring.

joeblubaugh:
`NumberKind` is just `FloatKind | IntKind`. If the scalar field in the Go type is a float, it can hold the value, and we shouldn't error.

Or at least that's my assertion :-) I don't know what peril this may cause, but it solved a BindType panic that seemed invalid.

### DIFF
--- a/schema_test.go
+++ b/schema_test.go
@@ -17,6 +17,7 @@ seqs: [
 			{
 				astring?: string
 				anint:   int64 | *42
+				afloat:  float64
 				abool:   bool
 			}
 		]
@@ -27,6 +28,7 @@ seqs: [
 type TestType struct {
 	Astring *string `json:"astring"`
 	Anint   int64   `json:"anint"`
+	Afloat  float64 `json:"afloat"`
 	Abool   bool    `json:"abool"`
 }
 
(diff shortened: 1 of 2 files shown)


joeblubaugh:
Digging into the cue subsume function, we run into this line in `core/adt/binop.go`:
```go
case leftKind&NumKind != 0 && rightKind&NumKind != 0:
	// n := c.newNum()
	return cmpTonode(c, op, c.Num(left, op).X.Cmp(&c.Num(right, op).X))
```

schema_test.go
```diff
@@ -27,6 +28,7 @@ seqs: [
 type TestType struct {
 	Astring *string `json:"astring"`
 	Anint   int64   `json:"anint"`
+	Afloat  float64 `json:"afloat"`
```
joeblubaugh, line 31: Adding this field causes a panic on `main`, but passes here.

sdboyer:
i ran into this - basically, `cue.FloatKind` doesn't seem to be used for value kind under the conditions i would have expected - when working on fixes to cue stdlib's openapi encoder. It may be a bug in core CUE, but it does appear to be the current behavior.

So, LGTM, as the test should now catch any potential change in behavior in CUE. thanks!

PR #112, opened 2023-02-21
Apply external functions at the end to avoid undesired results

spinillos:
We were applying the external functions at the beginning modifying the file prior to make the fixes. This approach complicates the fixes of the Go file inside `thema` and they should be at the end.

### DIFF
--- a/encoding/gocode/gen.go
+++ b/encoding/gocode/gen.go
@@ -81,8 +81,7 @@ func GenerateTypesOpenAPI(sch thema.Schema, cfg *TypeConfigOpenAPI) ([]byte, err
 		applyFuncs = append(applyFuncs, fixTODOComments())
 	}
 
-	for _, fn := range cfg.ApplyFuncs {
-		applyFuncs = append(applyFuncs, fn)
+	applyFuncs = append(applyFuncs, cfg.ApplyFuncs...)
 	}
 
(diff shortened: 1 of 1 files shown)


encoding/gocode/gen.go
```diff
@@ -75,10 +75,14 @@ func GenerateTypesOpenAPI(sch thema.Schema, cfg *TypeConfigOpenAPI) ([]byte, err
...
-	cfg.ApplyFuncs = append(cfg.ApplyFuncs, depointer, fixRawData())
 
+	applyFuncs := []dstutil.ApplyFunc{depointer, fixRawData()}
 	if !cfg.UseGoDeclInComments {
-		cfg.ApplyFuncs = append(cfg.ApplyFuncs, fixTODOComments())
+		applyFuncs = append(applyFuncs, fixTODOComments())
+	}
+
+	for _, fn := range cfg.ApplyFuncs {
+		applyFuncs = append(applyFuncs, fn)
```
sdboyer, line 85: spread operator works here
```suggestion
	applyFuncs = append(applyFuncs, cfg.ApplyFuncs...)
```

sdboyer (approved):
i must've misread this code when we first merged, because i figured this was what we were already doing :) we definitely want to apply internal transforms before external ones - injected transforms should not even be aware that a "pre-internal transforms" state exists.  output prior to

PR #115, opened 2023-02-23
Fix missing underscores

spinillos:
I found missing underscores inside structs when I upgraded `thema`: https://github.com/grafana/grafana/pull/63629.

Also I added logic to iterate maps and arrays deeply if we could have these cases.

### DIFF
--- a/encoding/gocode/decorators.go
+++ b/encoding/gocode/decorators.go
@@ -182,25 +182,41 @@ func fixUnderscoreInTypeName() dstutil.ApplyFunc {
 		switch x := c.Node().(type) {
 		case *dst.GenDecl:
 			specs, isType := x.Specs[0].(*dst.TypeSpec)
-			if isType && strings.Contains(specs.Name.Name, "_") {
-				oldName := specs.Name.Name
-				specs.Name.Name = withoutUnderscore(specs.Name.Name)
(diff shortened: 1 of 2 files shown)


AgnesToulet (approved):
Looks good!

I don't know how important/urgent is the Grafana PR so I don't want to block it but I think it would be nice to merge the `iterateXWithUnderscores` functions with the `iterateX` ones, WDYT?

encoding/gocode/decorators.go
```diff
@@ -244,6 +266,8 @@ func iterateArray(a *dst.ArrayType, existingRawFields map[string]bool) {
 	case *dst.Ident:
 		if existingRawFields[mx.Name] {
 			mx.Name = setStar(mx) + "interface{}"
+		} else if strings.Contains(mx.Name, "_") {
+			mx.Name = withoutUnderscore(mx.Name)
```
AgnesToulet, line 270: The issue with these lines is that the `fixRawData` decorator now fix some underscores (but not all) and the `fixUnderscores` decorator fix some raw data 😅 I think we should keep them separated.

I was thinking about something like:
```
iterateArray(a *dst.ArrayType, fn func(i *dst.Ident)) {
   ...
} 
```

PR #117, opened 2023-03-02
Fix surface Less function and add tests

AgnesToulet:
Fix `func (sv SyntacticVersion) Less(osv SyntacticVersion) bool` and add some tests for it and for related schema functions. 

Also fix `LatestInMajor` function that returns an out of range error if the next major version doesn't exist.


sdboyer:
ooof. what dumb bugs I've made! thank you!

PR #123, opened 2023-03-31
Lineage: First version of IsAppendOnly

AgnesToulet:
This is a first basic implementation of `IsAppendOnly(oldLineage, newLineage Lineage)` that does string comparison. 

Long-term, we probably want a more complex version that is less strict (allow authorized changes).

### DIFF
--- a/internal/cuetil/equal.go
+++ b/internal/cuetil/equal.go
@@ -1,0 +1,52 @@
+package cuetil
+
+import (
+	"regexp"
+	"strings"
+
+	"cuelang.org/go/cue"
+	"cuelang.org/go/cue/format"
+)
+
+func toString(val cue.Value) (string, error) {
+	node := val.Syntax(
+		cue.All(),
+		cue.Raw(),
+		cue.Schema(),
+		cue.Definitions(true),
+		cue.Docs(true),
+		cue.Hidden(true),
+	)
+
+	bytes, err := format.Node(node)
+	if err != nil {
+		return "", err
+	}
+
+	regWhitespace := regexp.MustCompile("\\s+")
+	res := regWhitespace.ReplaceAllString(string(bytes), "")
+	
+	regDef := regexp.MustCompile(`{_#def_#def:(.*)}`)
+	matches := regDef.FindStringSubmatch(res)
+	if len(matches) > 1 {
+		res = matches[1]
+	} else {
+		res = strings.Replace(res, "_#def_#def:", "", 1)
+	}
+
+	return res, nil
+}
+
+func Equal(val1 cue.Value, val2 cue.Value) bool {
+	string1, err := toString(val1)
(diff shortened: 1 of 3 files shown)


lineage.go
```diff
@@ -167,6 +168,16 @@ func (lin *UnaryLineage) First() Schema {
 	return lin.allsch[0]
 }
 
+// All returns all Schemas in the lineage. Thema requires that all valid lineages
+// contain at least one schema, so this is guaranteed to contain at least one element.
+func (lin *UnaryLineage) All() []Schema {
+	schemas := make([]Schema, len(lin.allsch))
+	for i, s := range lin.allsch {
+		schemas[i] = s
+	}
```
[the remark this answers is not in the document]
AgnesToulet: `lin.allsch` is an array of `*UnarySchema` so copy can't be used here.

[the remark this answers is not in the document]
sdboyer: hah! never realized that `copy` didn't deep-copy pointer values. TIL

internal/cuetil/equal.go
```diff
@@ -0,0 +1,56 @@
...
+		res = matches[1]
+	} else {
+		// Else it's just prefixed with this
+		res = strings.Replace(res, "_#def_#def:", "", 1)
+	}
+
+	return res, nil
+}
+
+func Equal(val1 cue.Value, val2 cue.Value) bool {
```
sdboyer, line 44: Need comment for exported func

lineage.go
```diff
@@ -268,3 +279,23 @@ type unaryConvLineage[T Assignee] struct {
 func (lin *unaryConvLineage[T]) TypedSchema() TypedSchema[T] {
 	return lin.tsch
 }
+
```
sdboyer, line 282: Need comment for exported func. Probably worth noting in that comment that the comparison algorithm this currently uses is unsophisticated and subject to change

lineage.go
```diff
@@ -268,3 +279,23 @@ type unaryConvLineage[T Assignee] struct {
...
+func IsAppendOnly(oldLineage Lineage, newLineage Lineage) bool {
```
sdboyer, line 283: i think this is the signature we want. Even if we can't convey much in the error message right now, it'll be crucial that a failure here is able to provide line-level feedback on where the problem was.

```suggestion
func IsAppendOnly(old, new Lineage) error {
```

sdboyer (changes requested):
i realize this is still marked draft, but it does seem increasingly like we need to get some version of it as a prerequisite to other work.

IMO it's fine to merge this even if the comparison algorithm is flawed, so long as we're happy with the `IsAppendOnly` function signature.

AgnesToulet:
Closing in favor of https://github.com/grafana/thema/pull/130

PR #129, opened 2023-04-17
Fix oapi-codegen check

spinillos:
`Oapi-codegen` was forked inside the library and it has a bug that it isn't able to detect that a duplicate field is the same. Mainly because its checking two pointers with equal operator.

This was reported and a PR was created in their repository to fix that bug.


sdboyer (approved):
LGTM!

PR #130, opened 2023-04-27
Lineage: First version of IsAppendOnly

AgnesToulet:
This is a first basic implementation of `IsAppendOnly(oldLineage, newLineage Lineage)`.

It compares schemas by looking if A subsumes B and B subsumes A. It returns and error if a schema as been updated or deleted.


internal/cuetil/equal.go
```diff
@@ -0,0 +1,18 @@
+package cuetil
+
+import (
+	"cuelang.org/go/cue"
+)
+
+// Equal reports nil when the two cue values subsume each other or an error otherwise
+func Equal(val1 cue.Value, val2 cue.Value) error {
+	if err := val1.Subsume(val2, cue.Raw(), cue.Schema(), cue.Definitions(true), cue.All(), cue.Final()); err != nil {
```
sdboyer, line 9: Looking at the implementation of `Subsume` in cue stdlib, i think that including `cue.All()` and `cue.Definitions(true)` are no-ops. (i'm also pretty sure that if they weren't, `cue.All()` makes the other option unnecessary).

We haven't exactly been careful in our usage elsewhere - in fact i'm guessing you got this set of options from the `compat` package - but it's time we get more precise about this.

Let's introduce some test cases to actually exercise this logic.

sdboyer (changes requested):

sdboyer (approved):
Looks like a great first pass!

PR #131, opened 2023-04-28
fix(cli): include CUEPath in BindingConfig

mildwonkey:
The cue path (-p flag) was not being passed down to GenerateLineageBinding, so the lineage lookup was missing from generated bindings. This fixes that, adds the cue import (if cuepath is set) and finally wraps the lineage value in quotes (if not already present). Both `lineage gen gobindings  -p lineage` and   `lineage gen gobindings  -p "lineage" produce the expected outcome.


encoding/gocode/binding.tmpl
```diff
@@ -5,7 +5,8 @@ import (
 	"io/fs"
 	"path/filepath"
 
-    "cuelang.org/go/cue/build"
+	{{ if .CUEPath }}"cuelang.org/go/cue"{{ end }}
+	"cuelang.org/go/cue/build"
```
sdboyer, line 9: shouldn't be necessary? we have a call to goimports in the generator that eliminates unused imports if they're not needed

not a big deal, though

encoding/gocode/gen.go
```diff
@@ -289,7 +289,7 @@ func GenerateLineageBinding(lin thema.Lineage, cfg *BindingConfig) ([]byte, erro
 		CueModName:          cfg.CueModName,
 		LoadDir:             cfg.LoadDir,
 		EmbedPath:           cfg.EmbedPath,
-		CUEPath:             cfg.CuePath.String(),
+		CUEPath:             fmt.Sprintf("%q", cfg.CuePath.String()),
```
sdboyer, line 292: oh well now i'm just _sure_ this never worked and i didn't test it properly 🤦🏻 

i think it's counterintuitive to expect any caller to have to add quotes on a string, though. Instead, let's add the quotes around where it's actually used in the template.

sdboyer (changes requested):
Good catch, thank you! Just some small things

encoding/gocode/binding.tmpl
mildwonkey: Well, it wasn't _necessary_, but this is the message thema emits without that line:

```goimports added the following import statements to dashboard_binding_gen.go:
	"cuelang.org/go/cue"
Relying on goimports to find imports significantly slows down code generation. Either add these imports with an AST manipulation in cfg.ApplyFuncs, or set cfg.IgnoreDiscoveredImports to true```

File: encoding/gocode/binding.tmpl [hunk 3]
```diff
@@ -5,7 +5,8 @@ import (
 	"io/fs"
 	"path/filepath"
 
-    "cuelang.org/go/cue/build"
+	{{ if .CUEPath }}"cuelang.org/go/cue"{{ end }}
```
sdboyer, line 8: ...this is what i should have suggested 😄 
```suggestion
	"cuelang.org/go/cue"
```

encoding/gocode/gen.go
mildwonkey: Where else should I put that, then? Those are the variables used in the template, and with that change it works wether or not quotes were part of the lineage on the command line.

encoding/gocode/binding.tmpl
mildwonkey: That's not better - if we aren't setting the cue path, we're not adding a line with cue.Path, so that import will be removed.

sdboyer: right, this is more of a consistency nit than anything. There are other imports in that list - at least `embed` and `io/fs` - which are only conditionally necessary, as well. But they're not wrapped in a conditional.

The conclusion i came to after writing a few of these templates is that it's easier to maintain a template with the full set of imports that may be necessary and rely on `goimports` to remove unnecessary ones for particular cases, than to have the additional conditionals in the template file.

mildwonkey: Ok, I can do that. Can I remove the warning about relying on goimports, if that's now how expected?

encoding/gocode/gen.go
sdboyer: wow, this is like my third time misreading something you did in the last couple days. i need to slow down 😭 

you're right, i'm wrong. i thought this was in `cmd/thema`. Here, this is fine - we're just constructing the private var to pass in to the template. This isn't exposed to external callers of `GenerateLineageBinding()`.

cmd/thema/thema
sdboyer, line 1: oop, also need to remove the committed binary

i'd ask that this be rebased out so that it's not in the git commit history 🙏🏻

sdboyer: i just pushed a commit to main to add this to gitignore

encoding/gocode/binding.tmpl
mildwonkey: And yet, when I ran the command, I got a warning, which is why I made the change. I'd like to either add the import or remove the warning.

Issue #121, opened 2023-03-12
Attach Thema to unity


sdboyer:
It's long overdue that we register Thema as part of unity - https://github.com/cue-unity/unity.

i've struggled with doing so so far because Thema has generally lacked native CUE tests. #82 starts turning the corner on that, so once that's merged, we can look at doing this.

PR #141, opened 2023-05-10
thema: Fix Translate() to same version

joanlopez:
It fixes the `Translate()` operation [when from and to versions are the same](https://github.com/grafana/thema/blob/main/translate.cue#L128), because the current implementation produces an error in such case:

```
_objold.steps: conflicting values [...#TranslatedInstance] and {} (mismatched types list and struct):
    ./testdata/internal/schmoo.cue:55:10
    ./translate.cue:37:10
    ./translate.cue:45:10
    ./translate.cue:128:16
```

It depends on https://github.com/grafana/thema/pull/139

Some open questions inline.

### DIFF
--- a/instance_test.go
+++ b/instance_test.go
@@ -1,0 +1,61 @@
+package thema_test
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/grafana/thema"
+	"github.com/grafana/thema/internal/txtartest/bindlin"
+	"github.com/grafana/thema/internal/txtartest/vanilla"
+	"github.com/grafana/thema/vmux"
+
+	"cuelang.org/go/cue/cuecontext"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInstance_Translate(t *testing.T) {
+	test := vanilla.TxTarTest{
+		Root: "./testdata/lineage",
+		Name: "core/instance/translate",
+	}
+
+	ctx := cuecontext.New()
+	rt := thema.NewRuntime(ctx)
+
+	test.Run(t, func(tc *vanilla.Test) {
+		if !tc.HasTag("multiversion") {
+			return
+		}
+
+		tName := strings.Replace(tc.Name(), t.Name()+"/", "", -1)
+
+		lin, lerr := bindlin.BindTxtarLineage(tc, rt)
+		require.NoError(tc, lerr)
+
+		for sch := lin.First(); sch != nil; sch = sch.Successor() {
+			for name, ex := range sch.Examples() {
+				for sch := lin.First(); sch != nil; sch = sch.Successor() {
+					// TODO: Validate lacunas
+					to := sch.Version()
+					tinst, _ := ex.Translate(to)
+					require.NotNil(t, tinst)
+
+					raw := tinst.Underlying()
+					require.True(t, raw.Exists())
+					require.NoError(t, raw.Err())
+
+					codec := vmux.NewJSONCodec("test")
+					rawBytes, err := codec.Encode(raw)
+					require.NoError(t, err)
+
+					wName := fmt.Sprintf("%s-%s-%s->%s.json", tName, name, ex.Schema().Version().String(), to.String())
+					w := tc.Writer(wName)
+					_, err = w.Write(rawBytes)
+					require.NoError(t, err)
+				}
+			}
+		}
+
+	})
+}
--- a/testdata/internal/basic.cue
+++ b/testdata/internal/basic.cue
@@ -121,7 +121,7 @@ basic: #Lineage & {
 			from: [1, 1]
 			input: _
 			result: {
-				renamed: input.init
+				renamed: input.renamed
 				all:     input.all
 				if (input.optional != _|_) {
 					optional: input.optional
--- a/testdata/lineage/basic-multiversion.txtar
+++ b/testdata/lineage/basic-multiversion.txtar
@@ -1,0 +1,337 @@
+# Basic schema evolution over time
+
+#multiversion
+-- in.cue --
+import "github.com/grafana/thema"
+
+thema.#Lineage
+name: "basic-multiversion"
+
+joinSchema: {
+    all: int32
+}
+
+schemas: [{
+    version: [0, 0]
+    schema: {
+        init: string
+    }
+    examples: {
+        simple: {
+            all:  42
+            init: "some string"
+        }
+    }
+},
+{
+    version: [0, 1]
+    schema: {
+        init:      string
+        optional?: int32
+    }
+    examples: {
+        withoutOptional: {
+            all:  42
+            init: "some string"
+        }
+        withOptional: {
+            all:  42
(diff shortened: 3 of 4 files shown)


instance_test.go
```diff
@@ -0,0 +1,61 @@
+package thema_test
```
joanlopez, line 1: Did the trick to avoid cyclic dependencies (_note it's using the `vmux` package, for instance, as well as some of `txtartest` packages that also rely on Thema's core_), but not sure how acceptable is it.

instance_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+		tName := strings.Replace(tc.Name(), t.Name()+"/", "", -1)
+
+		lin, lerr := bindlin.BindTxtarLineage(tc, rt)
+		require.NoError(tc, lerr)
+
+		for sch := lin.First(); sch != nil; sch = sch.Successor() {
+			for name, ex := range sch.Examples() {
+				for sch := lin.First(); sch != nil; sch = sch.Successor() {
+					// TODO: Validate lacunas
```
joanlopez, line 39: Any thought? Maybe I can just serialize them and write them to the file for the sake of regression.

instance_test.go
```diff
@@ -0,0 +1,61 @@
...
+		if !tc.HasTag("multiversion") {
+			return
+		}
+
+		tName := strings.Replace(tc.Name(), t.Name()+"/", "", -1)
+
+		lin, lerr := bindlin.BindTxtarLineage(tc, rt)
+		require.NoError(tc, lerr)
+
+		for sch := lin.First(); sch != nil; sch = sch.Successor() {
```
joanlopez, line 36: It takes a bit long (_locally it takes around ~50s_), but I think the exhaustiveness may worth it. Wdyt?

translate.cue
```diff
@@ -125,7 +125,7 @@ import "list"
 				list.Drop(_accum, 1)
 			},
 			// to version same as from version is a no-op
-			if cmp == 0 {},
+			if cmp == 0 {[]},
```
joanlopez, line 128: The actual fix! 🚀

testdata/internal/basic.cue
```diff
@@ -121,7 +121,7 @@ basic: #Lineage & {
 			from: [1, 1]
 			input: _
 			result: {
-				renamed: input.init
+				renamed: input.renamed
```
joanlopez, line 124: Small error caught while running the tests 😛

testdata/lineage/basic-multiversion.txtar
```diff
@@ -0,0 +1,337 @@
+# Basic schema evolution over time
```
joanlopez, line 1: As we discussed, I thought about re-using the one defined at `testdata/internal/basic.cue`, but there are some subtle differences, so I'm not sure if it's better to keep them for separate or try to bring examples there and try to re-use it. So far, I decided to start with the simpler.

testdata/lineage/basic-multiversion.txtar
```diff
@@ -0,0 +1,337 @@
...
+    from: [0, 3]
+    input: _
+    result: {
+        init: input.init
+        all:  input.all
+        if (input.optional != _|_) {
+            optional: input.optional
+        }
+        // TODO: withDefault: input.withDefault doesn't work
+        withDefault: "foo"
```
joanlopez, line 166: The use of `input.withDefault` does not only cause incompleteness on some of the test cases but also directly fails for some others. I guess we might want to explore that in a separate PR, but will probably require a more high-level discussion/decision.

instance_test.go
sdboyer: Yup, we shouldn't avoid testing what ought to be tested just because it's slow. We'll get to those perf issues soon enough

instance_test.go
sdboyer: Agreed, i think that's good

testdata/lineage/basic-multiversion.txtar
sdboyer: agreed, this is its whole own area for discussion

testdata/lineage/basic-multiversion.txtar
```diff
@@ -0,0 +1,337 @@
...
+{
+    to: [0, 0]
+    from: [0, 1]
+    input: _
+    result: {
+        init: input.init
+        all:  input.all
+    }
+}]
+-- out/core/instance/translate/lineage/basic-multiversion-simple-0.0->0.0.json --
```
sdboyer, line 206: i am delightfully surprised by how much MORE readable the whole set is than when then JSON is formatted and spread over multiple lines! i can actually visually scan this! 🎉

sdboyer: though...looking at it more, i'm finding that i kind of want to see both pre and post translation in the printed output, as otherwise i have to jump back and forth. WDYT?

instance_test.go
sdboyer: In general i do support use of this pattern. In this particular case, though, my gut feel is that the base thema package should probably avoid reliance on any other packages for helpers in tests. That helps us ensure we're layering our tests effectively, and keeps test failure signals cleaner (i.e. `github.com/grafana/thema` tests cannot fail as a result of any refactoring or bugs in `gh/grafana/thema/vmux` package)

i wouldn't say this is a hard rule, but where the implementations in other packages are fairly trivial, i think it's preferable to copy over importing. In this case:

* `bindlin` already has an [unexported equivalent](https://github.com/grafana/thema/blob/53a6fb01cf3135287b2501a08b9381b5c9831bbd/lineage_test.go#L82) in thema pkg tests
* the JSON codec in `vmux` is a pretty trivial wrapper around calls into other CUE funcs. i'd say it's fine to make an unexported helper in tests, same as with the bindlin func

testdata/lineage/basic-multiversion.txtar
sdboyer: i think it's fine to keep it simpler/separate - `testdata/internal/basic.cue` was intended to be part of a pure CUE corpus of tests, which is probably still worth having. (We could actually spit one out from the main corpus, then use e.g. `cue eval` and/or invoke the `thema` binary. That's also probably what we'll want to attach to unity - #121)

But for now, that doesn't exist, so making this part of the main corpus is 👍🏻

instance_test.go
joanlopez: Absolutely agree 💯 

> In general i do support use of this pattern. In this particular case, though, my gut feel is that the base thema package should probably avoid reliance on any other packages for helpers in tests. That helps us ensure we're layering our tests effectively, and keeps test failure signals cleaner (i.e. github.com/grafana/thema tests cannot fail as a result of any refactoring or bugs in gh/grafana/thema/vmux package)

testdata/lineage/basic-multiversion.txtar
joanlopez: Sounds good, let's try!

PR #144, opened 2023-05-11
thema: Fix Translate() to same version

joanlopez:
_(Descendant of https://github.com/grafana/thema/pull/141)_

It fixes the `Translate()` operation [when from and to versions are the same](https://github.com/grafana/thema/blob/main/translate.cue#L128), because the current implementation produces an error in such case:

```
_objold.steps: conflicting values [...#TranslatedInstance] and {} (mismatched types list and struct):
    ./testdata/internal/schmoo.cue:55:10
    ./translate.cue:37:10
    ./translate.cue:45:10
    ./translate.cue:128:16
```

### DIFF
--- a/testdata/lineage/basic-multiversion.txtar
+++ b/testdata/lineage/basic-multiversion.txtar
@@ -450,3 +450,1005 @@ lenses: [{
 Schema count: 6
 Schema versions: 0.0, 0.1, 0.2, 0.3, 1.0, 1.1
 Lenses count: 6
+-- out/encoding/gocode/TestGenerate/nilcfg --
+== basicmultiversion_type_0.0_gen.go
+package basicmultiversion
+
+// Basicmultiversion defines model for basicmultiversion.
+type Basicmultiversion struct {
+	Init string `json:"init"`
+}
+== basicmultiversion_type_0.1_gen.go
+package basicmultiversion
+
+// Basicmultiversion defines model for basicmultiversion.
+type Basicmultiversion struct {
+	Init     string `json:"init"`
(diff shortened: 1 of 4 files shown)


testdata/lineage/basic-multiversion.txtar
```diff
@@ -0,0 +1,452 @@
...
+{"init":"some string"}
+[]
+-- out/core/instance/translate/TestInstance_Translate/lineage/basic-multiversion-0.0-simple->0.1.json --
+{"init":"some string"}
+{"init":"some string"}
+[]
+-- out/core/instance/translate/TestInstance_Translate/lineage/basic-multiversion-0.0-simple->0.2.json --
+{"init":"some string"}
+{"init":"some string","withDefault":"foo"}
+[]
```
joanlopez, line 196: Does it look more clear now? 
It keeps the "one-liner idea" but writing all of them:
- From (example)
- To (translation result)
- Lacunas

joanlopez (commented):
Hey @sdboyer, sorry for the noise! I expected https://github.com/grafana/thema/pull/141 target branch to be updated to `main` as soon as https://github.com/grafana/thema/pull/139 got merged, but that haven't been the case.

Now most (or all) of the discussions from https://github.com/grafana/thema/pull/141 should be solved. Thanks!

testdata/lineage/basic-multiversion.txtar
sdboyer: yes, this is :chef-kiss:

sdboyer (approved):
looks great. In we go!

Issue #44, opened 2022-04-24
Make Thema errors excellent: clear, actionable, informative


sdboyer:
For anyone not terribly interested in learning about the details of CUE and just wanting Thema to be a schema framework (our core target use case), validation error output is nearly indecipherable for anything outside of trivial cases. I laid out some examples in this grafana issue: grafana/grafana#37859

While there remain underlying things that CUE itself probably ought to fix - in particular, eliminating irrelevant disjunction branches in the presence of a discriminator field - we can actually probably do most or all of that directly in Thema. When it comes to generating comprehensible error messages, our job will always be easier, because we know enough about our inputs to be able to treat one input as schema, and the other as data compared against schema.

This can be done in stages. First will be just getting the basic scalar/enum-type validation failures improved. After that, we can layer on handling for more complex structures, like complex struct disjunctions.

Note that #8 has the potential to make those disjunction easier - if discimrinator fields are a property we can discern from the way that composition itself is declared, then we have all we need to do branch elimination.

- [x] #45 
- [x] Actually write some test suites, yikes
- [ ] Handle disjunctions

frossano-grafana:
this is a usability issues, not a stability issue

sdboyer:
#82 begins the use of github.com/cockroachdb/errors. However, there's a fair bit of planning and design on its own that needs to go into adopting the library - see [this whole design doc](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20190318_error_handling.md#Compatibility-and-differences-with-Go-2113-error-semantics) that covers the behaviors.

i could be talked out of it, but i think for the role we would like to see Thema occupy - a highly portable utility that can tuck into individual programs or complex operational topologies - ensuring our errors have the properties outlined in that document's [motivation section](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20190318_error_handling.md#motivation-for-a-new-error-library-summary) seems like the right move. i'm fine with a complex errors implementation inside thema, so long as it's simple to use on the outside.

frossano-grafana:
I agree there is a fair bit of planning and design to do, and this is one of the things I talked abut with all the designers we interviewed. Error handling is both a technical and a usability issue, and we need to start from what we want to expose to final users and walk back to what we can achieve based on constraints in the implementation (there might be none, but there might be some big ones) 
Motivation behind the errors is super important, and for me in the end it boils down to: what do we want users to do when a specific error (or class of errors) comes out?

[...]

PR #153, opened 2023-05-17
thema: Add compat checking cases, improve errors

sdboyer:
This PR separates out a clear pattern within the `invalidlineages` testdata corpus for checking backwards compatibility invariants on schemas. It accomplishes this through negation - defining a series of schemas that violate a particular rule, and expecting that `BindLineage` throws an error when encountering such a case. This approach is sufficient to establish the rule, because the [law of the excluded middle](https://en.wikipedia.org/wiki/Law_of_excluded_middle) applies in this case - every schema relation is either backwards compatible, or not, and every syntactic version numbering either adheres to that, or not.

As indicated by several of these cases being in the `ToDo` list for `TestInvalidLineages`, Thema's checkers do not yet catch all of these cases.

The error output is also somewhat improved here, though we are not yet doing a reliably good job of providing line-level feedback in the way that we really ought to according to the principles outlined in #44.

Finally, this also fixes a huge 🤦🏻 - when doing backwards compatibility checking, the parameters being passed to `compat.ThemaCompatible()` were reversed, resulting in checking if e.g. 0.0 was backwards compatible with 0.1, rather than vice-versa.

cc @joanlopez @mildwonkey


joanlopez:
Looks like there's a test failing:

```
=== RUN   TestInstance_Translate/lineage/basic-multiversion
    instance_test.go:32: 
        	Error Trace:	/home/runner/work/thema/thema/instance_test.go:32
        	            				/home/runner/work/thema/thema/vanilla_txtar.go:387
        	Error:      	Received unexpected error:
        	            	schema 0.2 is not backwards compatible with schema 0.1:
        	            	(1) forced error mark
        	            	  | "invalid lineage"
        	            	  | github.com/cockroachdb/errors/withstack/*withstack.withStack::
        	            	Wraps: (2) schema 0.2 is not backwards compatible with schema 0.1:
        	            	  | required field is optional in subsumed value: withDefault
        	            	  | value not an instance
        	            	Error types: (1) *markers.withMark (2) *errors.errorString
```

But so far I don't get why 🤔 
Isn't it just [adding a new field](https://github.com/grafana/thema/blob/main/testdata/lineage/basic-multiversion.txtar#L43)?

PR #154, opened 2023-05-17
Allow unify references

spinillos:
When we extend an interface with another interface, the reference is inside `allOf` tag of open-api schema. 

Knowing that, we can move these values into a [properties](https://github.com/grafana/thema/compare/fix-unify-refs?expand=1#diff-b9558d57be0a327c45c7da2a7c5003369f13fd17ec6a2e0ca3d9c5ff32e730e2R268-R275) to render the interfaces instead their inner values. For any reason, if we don't have references, exists a default empty one that we needed to skip [here](https://github.com/grafana/thema/compare/fix-unify-refs?expand=1#diff-b9558d57be0a327c45c7da2a7c5003369f13fd17ec6a2e0ca3d9c5ff32e730e2R260).

Also, the current library implementation forces to set a type and json tag to every element and its the reason of the rest of the changes in this PR.

When we set group or expandref we aren't unable to find the reference of the interfaces and we have to put the inner interface values instead: [here](https://github.com/grafana/thema/compare/fix-unify-refs?expand=1#diff-6c3d5536cc36acb091426f4d6ca31ffde0bfada5d4d21ab557f35ec66794e9e3R71) and [here](https://github.com/grafana/thema/compare/fix-unify-refs?expand=1#diff-6c3d5536cc36acb091426f4d6ca31ffde0bfada5d4d21ab557f35ec66794e9e3R333).

**Update**: We have a known issue with new flattern pattern where we are unable to find some references when we use `_#schema` definition. Using _join and schema individually fixes the issue.


spinillos:
Each datasource is generating its own DataQuery struct... I understand that its something expected.

sdboyer:
> Each datasource is generating its own DataQuery struct... I understand that its something expected.

Yes, this is expected, at least for now. We'll soon need to deal with controls that allow us to generate these as references to imported types, similar to what we do in TS.

sdboyer (approved):
Shows some oddities we'll need to deal with in future issues, but i think this is good to go as it currently stands.

PR #157, opened 2023-05-18
thema: Try to return a concrete result from #Translate

### LINKED ISSUE #155: bug: inconsistently-shaped `instances` from `Translate` and `Validate`
This is very hard to describe in prose, so here's an example!
https://github.com/mildwonkey/thema-examples/tree/main/translate

Here's the incredibly simplified, pseudocode tl;dr:

```go
i0, _ := sch.Validate(exdata)
title, _ := i0.Underlying().LookupPath(cue.ParsePath("title")).String()
fmt.Println(title) // "foo"

i1, _ := i0.Translate(thema.SV(0, 1))
title1, err := i1.Underlying().LookupPath(cue.ParsePath("title")).String()
fmt.Println(err) // #Translate.out.result.result: field not found: title
```

The `title` field is unchanged and should be accessible with the same `Underlying().LookupPath` method.

joanlopez:
It adds a call to `Default()` after calling the CUE `#Translate` to try to return a concrete result, and deal with inconsistently-shaped instances from `Translate` and `Validate`, as described in #155.

It also adds a regression test specifically checking what's described on the issue. 

Fixes #155

### DIFF
--- a/instance.go
+++ b/instance.go
@@ -174,8 +174,10 @@ func (i *Instance) Translate(to SyntacticVersion) (*Instance, TranslationLacunas
 	lac := make(multiTranslationLacunas, 0)
 	out.LookupPath(cue.MakePath(cue.Str("lacunas"))).Decode(&lac)
 
+	raw, _ := out.LookupPath(cue.MakePath(cue.Str("result"), cue.Str("result"))).Default()
+
 	return &Instance{
-		raw:  out.LookupPath(cue.MakePath(cue.Str("result"), cue.Str("result"))),
+		raw:  raw,
 		name: i.name,
 		sch:  newsch,
 	}, lac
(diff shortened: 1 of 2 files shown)


instance.go
```diff
@@ -174,8 +174,10 @@ func (i *Instance) Translate(to SyntacticVersion) (*Instance, TranslationLacunas
 	lac := make(multiTranslationLacunas, 0)
 	out.LookupPath(cue.MakePath(cue.Str("lacunas"))).Decode(&lac)
 
+	raw, _ := out.LookupPath(cue.MakePath(cue.Str("result"), cue.Str("result"))).Default()
```
sdboyer, line 177: this is...slightly bizarre, let's make sure to add a comment indicating why the `Default()` call is there. Especially given that this is tightly coupled to the current implementation approach for `#Translate`

joanlopez: Sure! Comment added!

sdboyer (approved):
excellent, LGTM!

PR #158, opened 2023-05-18
thema: Fix undesired variable shadow on Instance.Translate test

joanlopez:
It slightly modifies the `TestInstance_Translate` test to stop shadowing the `sch` variable used for both, the _src_ and _dst_ of the `Translate` operation, which is behaving as expected but might cause issues and/or misunderstandings in the feature.


sdboyer (approved):
nice catch, and i appreciate the tidiness of a separate PR

PR #163, opened 2023-05-23
Example to attach Thema to Unity

joanlopez:
Example for #121:
- It defines the manifest as described at https://github.com/cue-unity/unity:
  - Look at `cue.mod/tests/tests.cue`.
- Adds a `foo.cue` file as an example of "test" (some CUE code to evaluate).
- Adds a `foo.txt` as a `testscript` definition, that `cue eval` the `foo.cue` file and compares the output.
- Adds a GitHub Action workflow to run `unity test` on the repository (project mode).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,13 +1,10 @@
+name: CUE Unity Test
+
 on:
   push:
     branches:
-      - main
-    tags:
-      - v*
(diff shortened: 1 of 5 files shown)


sdboyer:
awesome, this general approach looks great to me. Will be really, really helpful to get attached to unity

Issue #137, opened 2023-05-06
Improve performance of `#Lineage` and `BindLineage`


sdboyer:
Performance of lineage loading and binding declined significantly with lineage flattening (#82). Preliminary profiling (I've attached a - lol - screenshot) indicates that most of the problem occurs during calls to [`cue.Context.BuildInstance()`](https://pkg.go.dev/cuelang.org/go/cue#Context.BuildInstance), which indicates that it's the CUE evaluator struggling with working out all the new internal structures within the native CUE definition of `#Lineage`.

![Image](https://user-images.githubusercontent.com/21599/236626665-4e2e61b6-3c62-484a-b800-9bd41f30ef94.png)

(Note that the profiler has erroneously put `cue.Value.Unify` under `load.Instances` - the main cost there is incurred during the `BuildInstance()` call).)

My guess is that we have some low-hanging fruit here related to some of the new disjunctions introduced. I suspect we can express the same constraints in a different way and improve performance here ocnsiderably.

sdboyer:
We'll definitely want to add some benchmark tests as part of this. And i see that https://pkg.go.dev/golang.org/x/perf/cmd/benchstat exists, which could be useful in helping us track progress against this over time.

joanlopez:
> We'll definitely want to add some benchmark tests as part of this. And i see that https://pkg.go.dev/golang.org/x/perf/cmd/benchstat exists, which could be useful in helping us track progress against this over time.

Related with that, I've just discovered [gobenchdata](https://github.com/bobheadxi/gobenchdata), and it looks really nice!

sdboyer:
oh amazing, that looks like the path of least resistance

joanlopez:
Some initial insights from `BindLineage`: looks like most of the time is spent on [this line](https://github.com/grafana/thema/blob/main/lineage.go#L64).

sdboyer:
My guess is that `AppendSplit` is still just a symptom of an underlying structure that the evaluator is tripping over - it's the call to `cue.Value.Expr()` that's expensive, as this bit of the profile shows:

<img width="599" alt="Screenshot 2023-06-07 at 5 33 52 AM" src="https://github.com/grafana/thema/assets/21599/656cfa35-8350-427c-920d-28a8a5f39791">

So it's still looking like something in the pure CUE definitions that's causing the basic churn. But having the different benchmark cases does also help us differentiate a bit...lots of good info to go through.

PR #168, opened 2023-06-07
thema: Introduce basic benchmarks on Thema operations

sdboyer:
This PR introduces a basic set of benchmarks for certain Thema operations. We could probably stand to add a few more, but these should give us a basic baseline for comparison as we iterate on #137 and other performance issues.


joanlopez (approved):
LGTM!

PR #169, opened 2023-06-07
Lineage: Fix intermediate variable calc

joanlopez:
As reported by @mildwonkey and discussed [here](https://raintank-corp.slack.com/archives/C03BH67DS2J/p1686064633700929) and [here](https://raintank-corp.slack.com/archives/C03BH67DS2J/p1686067192429709?thread_ts=1686049601.258189&cid=C03BH67DS2J), the `Translate` operation panics with `index out of range [2] with length 2` when there's more than two majors (for instance `0.0`, `1.0` and `2.0`) in the lineage.

So, after some debugging, I realized the calc of the intermediate variable `_basis` was wrong, because as stated in the comment, it _`keeps the index of the first schema in each major version within the overall canonical schema sort ordering`_, so for example, with versions: `0.0`, `0.1`, `0.2`, `0.3`,`1.0`, `1.1` and `2.0`, I'd expect it to be `[0, 4, 6]` but it was used to be `[0, 4`], causing the aforementioned panic.

So, I tweaked it a bit to get what I guess it's the expected result.
Also added a test case that covers it, following Red-Green-Refactor!

This should make [this example](https://github.com/mildwonkey/thema-examples/tree/main/sch3ma) provided by @mildwonkey to work.

### DIFF
--- a/lineage.cue
+++ b/lineage.cue
@@ -165,7 +165,7 @@ import (
 	// _basis keeps the index of the first schema in each major version
 	// within the overall canonical schema sort ordering. This allows trivial
 	// schema retrieval from a syntactic version.
-	_basis: [0, for maj, _ in list.Drop(_counts, len(_counts)-1) {
+	_basis: [0, for maj, _ in list.Drop(_counts, 1) {
 		list.Sum(list.Take(_counts, maj+1))
 	}]
 
--- a/testdata/lineage/basic-multiversion.txtar
+++ b/testdata/lineage/basic-multiversion.txtar
@@ -601,9 +601,9 @@ lenses: [{
 {"toObj":{"init":"some string"},"optional":32,"withDefault":"bing"}
 []
 -- out/bind --
-Schema count: 6
-Schema versions: 0.0, 0.1, 0.2, 0.3, 1.0, 1.1
-Lenses count: 6
+Schema count: 7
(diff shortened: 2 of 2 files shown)


lineage.cue
```diff
@@ -165,7 +165,7 @@ import (
 	// _basis keeps the index of the first schema in each major version
 	// within the overall canonical schema sort ordering. This allows trivial
 	// schema retrieval from a syntactic version.
-	_basis: [0, for maj, _ in list.Drop(_counts, len(_counts)-1) {
+	_basis: [0, for maj, _ in list.Drop(_counts, 1) {
 		list.Sum(list.Take(_counts, maj+1))
 	}]
```
joanlopez, line 170:
```suggestion
	_basis: [0, for i, sch in list.Drop(SS, 1) if SS[i].version[0] < sch.version[0] {i + 1}]
```

Looking at the code around, looks to me that the `_pos` variable defined [here](https://github.com/grafana/thema/blob/main/lineage.cue#L148) (_as in the suggestion_) does exactly the same, in a way that imho it's even easier to read & understand.

So, if my assumption is correct, I'd suggest to define it that way, and we might also be able to reuse it and get rid of one variable / calc. Wdyt? 🤔

joanlopez:
I need to update some golden or tweak something I missed, but "unrelated" to the change, I'll do it later but feel free to review it meanwhile.

sdboyer (approved):
good catch, thank you!

Issue #138, opened 2023-05-08
encoding/gocode: Generate `any` instead of `interface{}`


sdboyer:
Now that `any` exists as a type in Go, we should generate it instead of `interface{}`.

I actually did this as part of #82, then rolled it back in e72fbab91f7a56ca02803ef70cd5e06cff931098 in order to help minimize the noise in the PR integrating #82 into grafana/grafana. Once that's done, doing this ticket should be as simple as reverting that commit.

joanlopez:
Moved back to "Ready to start" because I guess it's no longer blocked (the flattening PR has been merged) and it could be a quick win for anyone open to pick it up!

PR #175, opened 2023-06-13
Chore: Generate `any` instead of `interface{}` in Go code

undef1nd:
Does what it says in the title.

Fixes: https://github.com/grafana/thema/issues/138

### DIFF
--- a/testdata/lineage/any.txtar
+++ b/testdata/lineage/any.txtar
@@ -1,0 +1,269 @@
+# lineage containing fields that must generate any type in Go
+-- in.cue --
+import "github.com/grafana/thema"
+
+thema.#Lineage
+name: "any"
+schemas: [{
+    version: [0, 0]
+    schema: {
+        value: string | bool
+        optional?: string | bool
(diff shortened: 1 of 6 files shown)


sdboyer:
Seems like we don't have any test cases that exercise this. Could we add one to the txtar corpus, or modify an existing corpus case where field types don't matter so much?

testdata/lineage/any.txtar
```diff
@@ -0,0 +1,269 @@
+# lineage containing fields that must generate any type in Go
+-- in.cue --
+import "github.com/grafana/thema"
+
+thema.#Lineage
+name: "any"
```
sdboyer, line 6: Naming nitpicking 😄 

should also change the filename to match

```suggestion
name: "go-any"
```

sdboyer (changes requested):

sdboyer (approved):

sdboyer:
Thanks!

Now we need a quick follow-up PR against kindsys and then core Grafana to get this into place.

PR #176, opened 2023-06-26
thema: Tidy up base package and docs

sdboyer:
Assorted bits of housekeeping, related to thema readiness goals.

* Adds a bunch of missing godoc
* Gets rid of some unused code
* Adds internal checks on `thema.Instance` to ensure that it panics if not created by the system itself

### DIFF
--- a/assignable.go
+++ b/assignable.go
@@ -15,7 +15,7 @@ import (
 // If the provided T is a pointer, it will be dereferenced before verification.
 // Double pointers (or any n-pointer > 1) are not allowed.
 //
-// The provided T must struct-kinded, as it is a requirement that all Thema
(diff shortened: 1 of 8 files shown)


assignable.go
```diff
@@ -15,8 +15,8 @@ import (
 // If the provided T is a pointer, it will be dereferenced before verification.
 // Double pointers (or any n-pointer > 1) are not allowed.
 //
-// The provided T must necessarily be of struct type, as it is a requirement
-// that all Thema schemas are of base type struct.
+// The provided T must struct-kinded, as it is a requirement that all Thema
```
joanlopez, line 18:
```suggestion
// The provided T must be struct-kinded, as it is a requirement that all Thema
```

undef1nd (approved):

PR #177, opened 2023-06-27
Comment Pick and PickDef to make generation faster

spinillos:
These 2 functions make the generation slower and in case of changes in [this PR](https://github.com/grafana/grafana/pull/62142/files#diff-abdfd6a9318ae27ef245c1da7557b327ca1e4e6a10ee220b76e162c6865eeb28R344), it hangs forever...

The dashboard generation in my local took 1min with 27sec.

Looks like that at the moment we aren't use them (?), so I commented them to unblock the dashboard changes.

### DIFF
--- a/lineage.cue
+++ b/lineage.cue
@@ -9,7 +9,7 @@ import (
 // evolutionary history of a particular kind of object: every schema that has
 // ever existed for that object, and the lenses that allow translating between
 // those schema versions.
-#Lineage: L={
+#Lineage: {
 	// The name of the thing specified by the schemas in this lineage.
 	//
 	// A lineage's name must not change as it evolves.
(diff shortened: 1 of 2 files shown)


lineage.cue
```diff
@@ -92,8 +92,8 @@ import (
 		_schemas: [#SchemaDef & {version: [0, 0]}]
 	}
 
-	_forwardLenses: [ for lens in SL if {lens.to[0] > lens.from[0]} {lens}]
-	_backwardLenses: [ for lens in SL if {lens.to[0] <= lens.from[0]} {lens}]
+	_forwardLenses: [ for lens in lenses if {lens.to[0] > lens.from[0]} {lens}]
+	_backwardLenses: [ for lens in lenses if {lens.to[0] <= lens.from[0]} {lens}]
```
undef1nd, line 96: @sdboyer Do I get it right that this may be removed if the order check will be in Go code now?

spinillos: I only removed the alias from [here](https://github.com/grafana/thema/pull/177/files#diff-331e4a81b037b5b35f4015d6aa015ed24f7d301d8e0491660a8aa3397f89d6b5L82), so it should be the same.

sdboyer: these are needed - we're constructing values there which are used in `#Translate`

joanlopez: I guess it will depend on how restrictive we define the requirements for lenses definition

sdboyer (approved):
LGTM, ty! we can re-enable these later

Issue #164, opened 2023-05-23
Create standard error models for all thema operations


sdboyer:
As a system that deals with validation, error design is an essential part of Thema's UX. We need a clear, consistent model for the errors that Thema will produce, including clear and easily introspected Go error types.

Each major thema operation (validation, translation) needs its own error model. Lineage binding also needs an error model - probably more urgently than either validate or translate, as the lineage binding error model is basically the "lineage typechecker" that enforces all invariants.

Properties for thema's error models should include:

* A well-defined tree of error types, identifiable via `errors.Is`
* A clear aggregation mechanism for situations where multiple errors may occur (e.g. validation fails on multiple lines)
* Network portability, so that e.g. a standalone thema http server reliably produces the same information as would have arisen from something in-process

I got started on the mechanics of this in #82 by pulling in github.com/cockroachdb/errors, but ended up deferring it in the interest of actually the overall PR done.

This supersedes #44, as the scope of that issue was really just focused on validate errors - only one of the models we need. Still relevant is the discussion about noisy errors in disjunctions, though.

sdboyer:
This is gonna take some discussion before starting - probably needs a design doc.

sdboyer:
Note that this:

> Line-level information about errors

is deceptively tricky. On the one hand, CUE very _awesomely_ carries line-level traceability for validation errors by construction. But there are caveats.

Line-level error info ultimately ties back to associations between the vertices that underlie a [`cue.Value`](https://pkg.go.dev/cuelang.org/go@v0.5.0/cue#Value) and source file, specifically a [`token.Pos`](https://pkg.go.dev/cuelang.org/go@v0.5.0/cue/token#Pos). There are many things that can cause a `cue.Value` to lack such information - e.g., if it was created by `cue.Value.Unify()`, if it is the result of a comprehension.

It's _not_ as simple as "if `cue.Value.Source()` returns nil, there will be no line info in errors." That wouldn't simplify things much, anyway. But i do not know of a clean, general rule that clearly tells us in what circumstance we can expect [errors](https://pkg.go.dev/cuelang.org/go@v0.5.0/cue/errors) coming from to have line information (a non-empty return from [this](https://pkg.go.dev/cuelang.org/go@v0.5.0/cue/errors#Positions)).   (cc @myitcv)

[...]

PR #178, opened 2023-06-28
feat: Return sane errors from Translate

### LINKED ISSUE #167: Translate() may return an invalid instance
In [this example](https://github.com/mildwonkey/thema-examples/tree/main/datasource-change), schema 1.0 changes a string to a struct with a required inner field. Thema accepts a no-op lens, which results in Validate() returning an instance with that struct set to it's zero value, which then fails `Validate()`.

I'm not precisely sure what the fix is, here - there's more than one issue at play. Thema could complain about the no-op lens, but I'm not sure if that should be an error. Should thema validate that breaking schema changes are addressed in a lens or lacuna when generating code? Or is this something we could add to a thema test framework for developers to inspect their lenses?

I do think `Translate()` should return error if the translated instance fails validate (though we need to evaluate performance on Validate() before adding extra calls).

Feel free to close or modify this however you see fit!

sdboyer:
This PR introduces better error handling for `Translate`.

The function signature is remaining the same for `Translate`, given the aspiration to eventually have it not return errors (as described in the new docs for it). It will still panic if errors occur.

But, we also introduce `TranslateErr`, which returns errors instead of panicking. All errors which could emerge from `TranslateErr` are instances (`errors.Is`) of the new `ErrInvalidLens` sentinel error value. (That won't actually work yet because we're still figuring out how to use cockroachdb/errors in the context of #164, but once we do this should "Just Work")

The error sentinel disambiguation will be especially helpful in the context of vmuxers, where a non-nil error could _either_ come from the validate step, or from the translate step.

Fixes #167

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -67,7 +67,7 @@ var (
 )
 
 // Translation errors. These all occur as a result of an invalid lens. Currently
-// these may be returned from [thema.Instance.TranslateErr]. Eventually, it is
+// these may be returned from [thema.Instance.Translate]. Eventually, it is
 // hoped that they will be caught statically in [thema.BindLineage] and cannot
 // occur at runtime.
 var (
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -66,6 +66,27 @@ var (
 	ErrInvalidOutOfBounds = errors.New("data is out of schema bounds")
 )
 
+// Translation errors. These all occur as a result of an invalid lens. Currently
+// these may be returned from [thema.Instance.TranslateErr]. Eventually, it is
+// hoped that they will be caught statically in [thema.BindLineage] and cannot
+// occur at runtime.
+var (
+	// ErrInvalidLens indicates that a lens is not correctly written. It is the parent
+	// to all other lens and translation errors, and is a child of ErrInvalidLineage.
+	ErrInvalidLens = errors.New("lens is invalid")
+
+	// ErrLensIncomplete indicates that translating some valid data through
+	// a lens produced a non-concrete result. This always indicates a problem with the
+	// lens as it is written, and as such is a child of ErrInvalidLens.
+	ErrLensIncomplete = errors.New("result of lens translation is not concrete")
+
+	// ErrLensResultIsInvalidData indicates that translating some valid data through a
+	// lens produced a result that was not an instance of the target schema. This
+	// always indicates a problem with the lens as it is written, and as such is a
+	// child of ErrInvalidLens.
+	ErrLensResultIsInvalidData = errors.New("result of lens translation is not valid for target schema")
+)
+
 // Lower level general errors
 var (
 	// ErrValueNotExist indicates that a necessary CUE value did not exist.
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -35,17 +35,17 @@ type ValidationError struct {
 
 // Unwrap implements standard Go error unwrapping, relied on by errors.Is.
 //
-// All ValidationErrors wrap the general ErrNotAnInstance sentinel error.
+// All ValidationErrors wrap the general ErrInvalidData sentinel error.
 func (ve *ValidationError) Unwrap() error {
-	return ErrNotAnInstance
+	return ErrInvalidData
 }
 
 // Validation error codes/types
 var (
-	// ErrNotAnInstance is the general error that indicates some data failed validation
+	// ErrInvalidData is the general error that indicates some data failed validation
 	// against a Thema schema. Use it with errors.Is() to differentiate validation errors
 	// from other classes of failure.
-	ErrNotAnInstance = errors.New("data not a valid instance of schema")
+	ErrInvalidData = errors.New("data not a valid instance of schema")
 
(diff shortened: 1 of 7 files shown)


errors/errors.go
```diff
@@ -66,6 +66,27 @@ var (
 	ErrInvalidOutOfBounds = errors.New("data is out of schema bounds")
 )
 
+// Translation errors. These all occur as a result of an invalid lens. Currently
+// these may be returned from [thema.Instance.TranslateErr]. Eventually, it is
+// hoped that they will be caught statically in [thema.BindLineage] and cannot
+// occur at runtime.
```
joanlopez, line 72: Is that aspiration fair as soon as we foresee the need of adding support for "delegated" translations? Or would that be another, separated, API? 🤔 

NB: By "delegated" translations, I mean a translation that cannot be performed just with plain CUE but relies on external systems (like a DB) and/or requires another way to define it (in Go, for instance).

errors/errors.go
```diff
@@ -66,6 +66,27 @@ var (
...
+	// a lens produced a non-concrete result. This always indicates a problem with the
+	// lens as it is written, and as such is a child of ErrInvalidLens.
+	ErrLensIncomplete = errors.New("result of lens translation is not concrete")
+
+	// ErrLensResultIsInvalidData indicates that translating some valid data through a
+	// lens produced a result that was not an instance of the target schema. This
+	// always indicates a problem with the lens as it is written, and as such is a
+	// child of ErrInvalidLens.
+	ErrLensResultIsInvalidData = errors.New("result of lens translation is not valid for target schema")
+)
```
joanlopez, line 88: When docs says: `is a child of [error]`, what does that actually mean?

Does it mean it should be checkable with `errors.Is`? In such case, I cannot see where that "inheritance" (or better said: wrapping) is happening, but I might be missing something here.

sdboyer: > Does it mean it should be checkable with errors.Is?

Yeah, that's the intention. You're right though, "child" doesn't mean a clear thing here, and it should be phrased in a way where that's clear.

instance.go
```diff
@@ -167,7 +170,22 @@ func (inst *TypedInstance[T]) ValueP() T {
...
+//
+// Errors can only occur in Translate if lenses are not written correctly.
+// Eventually, it is hoped that all possible error cases on lenses will be
+// determinable statically, and turned into lineage validity errors that are
+// caught by [BindLineage]. If that goal is reached, this function will never
+// panic, and [Instance.TranslateErr] will be deprecated.
+func (i *Instance) Translate(to SyntacticVersion) (*Instance, TranslationLacunas) {
+	inst, lac, err := i.TranslateErr(to)
+	if err != nil {
+		panic(err)
```
spinillos, line 183: Could we remove this panic and return an error instead?

sdboyer: that's what `TranslateErr` is for - though maybe it would be better to have a `TranslateP` that panics and `TranslateErr` returns an error

joanlopez: I like this idea because of the fact of leaving the simple version of `Translate` open for future releases, optionally moving towards deprecations instead breaking changes - tho it's of course not critical because of the fact we're on `v0.x` 😄

errors/errors.go
joanlopez: > which would be lovely from a "keep thema's Go interface stable" perspective.

Perhaps adding functional options? I'd say it's quite idiomatic (fair?) and would leave us more flexibility.

> making that degraded mode a normal part of the contract means that a non-erroring Translate() is just permanently off the table

Right, yeah! 😅
But I'm curious, how do you think that'd change with lacunas? 
Would it be something like: errors vs "unsatisfied/unhandled" lacunas?

instance.go
```diff
@@ -187,7 +191,11 @@ func (inst *TypedInstance[T]) ValueP() T {
 // result in the exact original data. Input state preservation can be fully
 // achieved in the program depending on Thema, so we avoid introducing
 // complexity into Thema that is not essential for all use cases.
-func (i *Instance) Translate(to SyntacticVersion) (*Instance, TranslationLacunas) {
+//
+// Errors only occur in cases where lenses were written in an unexpected way -
+// for example, not all fields were mapped over, and the resulting object is not
+// concrete. All errors returned from this func will children of [terrors.ErrInvalidLens].
```
joanlopez, line 197:
```suggestion
// concrete. All errors returned from this func will be children of [terrors.ErrInvalidLens].
```

vmux/typed.go
```diff
@@ -48,7 +48,11 @@ func NewTypedMux[T thema.Assignee](sch thema.TypedSchema[T], dec Decoder) TypedM
 			}
 
 			if inst, ierr := isch.Validate(v); ierr == nil {
-				trinst, lac := inst.Translate(sch.Version())
+				trinst, lac, err := inst.Translate(sch.Version())
+				if err != nil {
+
```
joanlopez, line 53: Shouldn't we do something here?

instance.go
sdboyer: ahhhh whoops missed this, sorry

Issue #174, opened 2023-06-12
Enforce ordering of items in `schemas` and `lenses` during `BindLineage`


sdboyer:
#173 removed certain native CUE logic that sorted `lenses` and `schemas` in a lineage into a known order. To replace this, we need to add checks to `BindLineage()` that enforce ordering invariants on the raw lineage inputs.

sdboyer:
@undef1nd this is the one you were aiming to pick up, right?

undef1nd:
Yes, I think it's one of those errors handling improvements you suggested.

PR #179, opened 2023-06-29
Errors: Ensure lenses and schemas are ordered by version

undef1nd:
This PR does what the title says, solves https://github.com/grafana/thema/issues/174

I still need to figure out how to make cockroachdb/errors work with std errors package.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -82,6 +82,9 @@ var (
 	// by version.
 	ErrInvalidSchemasOrder = errors.New("schemas in lineage are not ordered by version")
 
+	// ErrInvalidLensesOrder indicates that lenses are in the wrong order - they must be sorted by `to`, then `from`.
+	ErrInvalidLensesOrder = errors.New("schemas in lineage are not ordered by version")
+
 	// ErrVersionNotExist indicates that no schema exists in a lineage with a
 	// given version.
 	ErrVersionNotExist = errors.New("lineage does not contain schema with version") // ErrNoSchemaWithVersion
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -1,6 +1,8 @@
 package errors
 
-import "github.com/cockroachdb/errors"
+import (
+	"github.com/cockroachdb/errors"
+)
 
 // ValidationCode represents different classes of validation errors that may
 // occur vs. concrete data inputs.
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -78,6 +78,10 @@ var (
 	// or more of the Thema invariants.
 	ErrInvalidLineage = errors.New("invalid lineage")
 
+	// ErrInvalidSchemasOrder indicates that schemas in a lineage are not ordered
+	// by version.
+	ErrInvalidSchemasOrder = errors.New("schemas in lineage are not ordered by version")
+
 	// ErrVersionNotExist indicates that no schema exists in a lineage with a
 	// given version.
 	ErrVersionNotExist = errors.New("lineage does not contain schema with version") // ErrNoSchemaWithVersion
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -85,7 +85,7 @@ var (
 	ErrInvalidSchemasOrder = errors.New("schemas in lineage are not ordered by version")
 
 	// ErrInvalidLensesOrder indicates that lenses are in the wrong order - they must be sorted by `to`, then `from`.
-	ErrInvalidLensesOrder = errors.New("schemas in lineage are not ordered by version")
+	ErrInvalidLensesOrder = errors.New("lenses in lineage are not ordered by version")
 
 	// ErrVersionNotExist indicates that no schema exists in a lineage with a
 	// given version.
--- a/lineage.cue
+++ b/lineage.cue
@@ -109,18 +109,6 @@ import (
 	//		}
 	//	}
 
-//	_schemasAreOrdered: [ for i, sch in SS {
-//		if i > 0 {
-//			[
-//				// sequence is ok if minor version bump
-//				if (sch.version[0] == SS[i-1].version[0]) && (sch.version[1] == SS[i-1].version[1]+1) {true},
-//				// or if major version bump, minor back at 0
-//				if (sch.version[0] == SS[i-1].version[0]+1) && (sch.version[1] == 0) {true},
-//				false,
-//			][0] & true
-//		}
-//	}]
-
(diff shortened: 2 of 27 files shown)


lineage.cue
```diff
@@ -109,17 +109,17 @@ import (
...
+//		if i > 0 {
+//			[
+//				// sequence is ok if minor version bump
+//				if (sch.version[0] == SS[i-1].version[0]) && (sch.version[1] == SS[i-1].version[1]+1) {true},
+//				// or if major version bump, minor back at 0
+//				if (sch.version[0] == SS[i-1].version[0]+1) && (sch.version[1] == 0) {true},
+//				false,
+//			][0] & true
+//		}
+//	}]
```
undef1nd, line 122: Is this still needed? I don't think it is

errors/errors.go
```diff
@@ -80,6 +80,13 @@ var (
 	// or more of the Thema invariants.
 	ErrInvalidLineage = errors.New("invalid lineage")
 
+	// ErrInvalidSchemasOrder indicates that schemas in a lineage are not ordered
+	// by version.
+	ErrInvalidSchemasOrder = errors.New("schemas in lineage are not ordered by version")
+
+	// ErrInvalidLensesOrder indicates that lenses are in the wrong order - they must be sorted by `to`, then `from`.
+	ErrInvalidLensesOrder = errors.New("schemas in lineage are not ordered by version")
```
spinillos, line 88: I understand that it should be like this:
```suggestion
	ErrInvalidLensesOrder = errors.New("lenses in schema are not ordered by version")
```

lineage.cue
spinillos: We are only check this here. So I think that we can delete it.

errors/errors.go
spinillos: Maybe this error could be structured and pass the version as argument.

lineage_test.go
```diff
@@ -32,9 +33,17 @@ func TestBindLineage(t *testing.T) {
 			t.Skip("case is tagged #slow, skipping for -short")
 		}
 
+		expected, ok := tc.Value("expectedErrorMessage")
+		assert.Equal(tc, ok, err != nil, "expected error message, but got none: %v", expected)
+
 		if err != nil {
+			if ok {
```
spinillos, line 40: Could be have `err != nil` and `ok == true` here? Looks like a bit odd.

lineage.cue
joanlopez: At the very least we'll need to figure out another way to have such information at CUE level without breaking, but if we don't find it useful (at least for now), we can get rid of it!

lineage_test.go
```diff
@@ -32,9 +33,17 @@ func TestBindLineage(t *testing.T) {
 			t.Skip("case is tagged #slow, skipping for -short")
 		}
 
+		expected, ok := tc.Value("expectedErrorMessage")
```
joanlopez, line 36: Do we plan to use `-- error --` file for such purpose? 🤔

undef1nd: Initially I planned to assert on the type of error rather than on the message, so used a txtar tag `expectedErrorMessage`. But as I didn't figure out how to properly use `errors.Is` with cockroach errors, I rely on the error message for now and have been discussing with @spinillos that I should use `-- error --` file instead of the tag for that. This seems to be the pattern used across the other tests.

>The problem I see with that is that we reuse the same .txtar files for multiple use cases (different tests) and there might be different error outputs. Any idea/thought?

Good question. Need to think about it, don't have any good ideas right now.

errors/errors.go
spinillos: Yes, you are right

undef1nd: @spinillos the combination of `from, to` versions for a lens is already passed here https://github.com/grafana/thema/pull/179/files#diff-aade326d3512b5a2ada6faa791ddec468f2a0adedb352339c9e314e74c8949d2R104 - it's done for both current and previous lenses.

testdata/invalidlineage/second-schema-versionless.txtar
```diff
@@ -31,9 +31,4 @@ lenses: [
 	}
 ]
 -- out/bindfail --
-_schemasAreOrdered.1: _schemasAreOrdered.1: conflicting values true and false:
-    ../../lineage.cue:113:3
-    ../../lineage.cue:114:4
-    ../../lineage.cue:119:5
-    ../../lineage.cue:120:11
-    ./in.cue:4:1
+schema 0.0 must be backwards incompatible with schema 0.0: introduce a breaking change, or redeclare as version 0.1
```
sdboyer, line 34: hah, this is a funny side effect of using `[0, 0]` as a default value in `#SchemaDef`

testdata/lineage/unordered-lenses.txtar
sdboyer, line 1: move to `testdata/invalidlineage`

testdata/lineage/unordered-schemas.txtar
sdboyer, line 1: move to `testdata/invalidlineage`

lineage_test.go
sdboyer: i've generally found the `-- error --` file to be a bit confusing. Like, i'm not sure what the standard use pattern is where you'd want to _not_ fail on an error. If it's an expected error, then it should be in the regular output for a particular test

in any case, i think our particular issues here will be solved by moving these txtar files into the `invalidlineages` testdata folder.

lineage_test.go
```diff
@@ -33,7 +33,8 @@ func TestBindLineage(t *testing.T) {
 		}
 
 		if err != nil {
-			tc.Fatalf("error binding lineage: %+v", err)
+			tc.ValidateErrorOrFail(err)
```
sdboyer, line 36: can get rid of this by just having a test where failure is the expected output - which is what should just happen by moving the fixtures over to `testdata/invalidlineage`

sdboyer (changes requested):
looks good! just need to move the fixtures over to the `invalidlineage` testdata dir, which should then make it possible to remove changes to `lineage_test.go` and the `ValidateErrorOrFail` err.

lineage.cue
undef1nd: @sdboyer any opinion reg this thing? thanks

sdboyer: 👍🏻

spinillos (approved):
Looks like nice

sdboyer (approved):

PR #183, opened 2023-07-12
First pass on refining error messages

### LINKED ISSUE #150: Bring validation error output in sync with lineage flattening
I forgot to update thema's validation error wrapping as part of #82. As a result, validation error output now contains output like that seen in #140:

```
<dashboard@v0.1>._sortedSchemas.1._#schema.gnetId
```

instead of what it's supposed to be:

```
<dashboard@v0.1>.gnetId
```

Fix should be trivial, but maybe best to do it only after we have some actual tests for the validation error wrapper 🤦🏻

joanlopez:
This is a first pass on refining error messages, with the following improvements:
- Hide CUE path details from validation error output, just display the failing attribute name (fixes #150) 
- Remove the spammy message, which looked weird for a library error
- Simplify the error message for well known numeric formats (`int64` instead of `int & >=-9223372036854775808 & <=9223372036854775807"`)
- Add a new case within `mungeValidateErr` to handle uncovered error case with msg `invalid value %v (out of bound %s)`

Thank you all!

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -8,6 +8,7 @@ import (
 	"cuelang.org/go/cue"
 	"cuelang.org/go/cue/errors"
 	"cuelang.org/go/cue/token"
+
 	terrors "github.com/grafana/thema/errors"
 )
 
@@ -168,9 +169,6 @@ func mungeValidateErr(err error, sch Schema) error {
 			errs = append(errs, err)
 			continue
 		}
-
-		// We missed a case, wrap CUE err in a plea for help
-		errs = append(errs, fmt.Errorf("no Thema handler for CUE error, please file an issue against github.com/grafana/thema\nto improve this error output!\n\n%w", ee))
 	}
 	return errs
 }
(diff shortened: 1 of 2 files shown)


K-Phoen (approved):

validate.go
```diff
@@ -152,6 +170,12 @@ func mungeValidateErr(err error, sch Schema) error {
 				break
 			}
 
+			if schkind == cue.IntKind || schkind == cue.FloatKind || schkind == cue.NumberKind {
```
sdboyer, line 173: i'd suggest doing a bitwise comparison here - i never quite trust that we get a value where it's _just_ one of these, so this covers the bases better

```suggestion
			if schkind & (cue.IntKind | cue.FloatKind | cue.NumberKind) != 0 {
```

sdboyer (changes requested):
just the one thing, then i think this is good to go. thanks everyone!

undef1nd (approved):

validate.go
joanlopez: Diving deeper, I just realized that `NumberKind Kind = IntKind | FloatKind`, so I guess we could just do: `schkind & cue.NumberKind`, which is equivalent, and even simpler, right? 🤔 cc/ @undef1nd @K-Phoen

sdboyer (approved):

PR #185, opened 2023-07-14
Test `Schema.Validate()` against the corpus

K-Phoen:
This PR adds a test case that relies on the existing corpus of lineages and schemas to test the validation process.

The idea is to extend the current corpus with examples of inputs to validate against, see the errors that come out, evaluate their correctness/friendliness, and then implement the necessary fixes.

Here we focused mainly on being able to run this type of test. A following PR will augment the corpus with more test cases, that we can rely on to safely refactor our "error munging" function.

### DIFF
--- a/internal/txtartest/vanilla/vanilla_txtar.go
+++ b/internal/txtartest/vanilla/vanilla_txtar.go
@@ -36,6 +36,7 @@ import (
 	"cuelang.org/go/pkg/encoding/yaml"
 	"github.com/google/go-cmp/cmp"
 	"github.com/grafana/thema/internal/envvars"
+	"github.com/grafana/thema/internal/util"
 	"golang.org/x/tools/txtar"
 )
 
@@ -45,6 +46,8 @@ import (
 type TxTarTest struct {
 	// Run TxTarTest on this directory.
 	Root string
+
+	ThemaFS fs.FS
 
 	// Name is a unique name for this test. The golden file for this test is
 	// derived from the out/<name> file in the .txtar file.
@@ -95,6 +98,8 @@ type Test struct {
 	outFiles []file
 
 	Archive *txtar.Archive
+
+	ThemaFS fs.FS
 
 	// The absolute path of the current test directory.
 	Dir string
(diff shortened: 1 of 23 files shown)


internal/txtartest/vanilla/vanilla_txtar.go
```diff
@@ -280,24 +285,31 @@ func formatVanillaNode(t *testing.T, n ast.Node) []byte {
...
-	return LoadVanilla(t.Archive, t.Dir, args...)
+	return LoadVanilla(t.ThemaFS, t.Archive, args...)
 }
 
 // LoadVanilla loads the instances of a txtar file. By default, it only loads
-// files in the root directory. Relative files in the archive are given an
-// absolute location by prefixing it with dir.
-func LoadVanilla(a *txtar.Archive, dir string, args ...string) []*build.Instance {
+// files in the root directory.
+func LoadVanilla(themaFS fs.FS, a *txtar.Archive, args ...string) []*build.Instance {
```
K-Phoen, line 293: **Note:** we changed the way CUE files are loaded for tests.

Errors include an absolute path to the file that caused it (both for the input data and the schema), implying that running the tests on a different machine would produce a different error message.
To work around this feature, we decided to load CUE files in a way that exclusively relies on a virtual filesystem.

undef1nd (approved):
We've got more tests now! 🎉

internal/txtartest/vanilla/vanilla_txtar.go
sdboyer: Excellent!

testdata/isappendonly/invalid/default.txtar
```diff
@@ -22,6 +22,6 @@ lin2: schemas: [{
 }]
 -- out/isappendonly-fail --
 field aunion not present in {aunion:*"bar" | "foo" | "baz"}:
-    ../../../lineage.cue:223:10
-    ./in.cue:16:13
+    ../../../../../../../../cue.mod/pkg/github.com/grafana/thema/lineage.cue:223:10
```
sdboyer, line 25: whew, it's a lot of dot-dots. not the end of the world, though

validate.go
```diff
@@ -106,6 +106,7 @@ func (vf validationFailure) Error() string {
 	return buf.String()
 }
 
+// HERE BE DRAGONS, BRING A SWORD.
```
sdboyer, line 109: ❤️ this so much

sdboyer (approved):
fantastic progress!

PR #187, opened 2023-07-21
thema: Initial pass at Go lenses/migrations

sdboyer:
This introduces an alternate mechanism for writing lenses: passing in a slice of `ImperativeLens` to `BindLineage`:

```go
type ImperativeLens struct {
	To, From SyntacticVersion
	Mapper   func(inst *Instance, to Schema) (*Instance, error)
}
```

where each function has exactly the same scope of responsibility as would otherwise be fulfilled by a declarative lens written in CUE.

The caller provides these via a new `BindOption` called `ImperativeLenses()`. In the current implementation, the caller must choose to implement their lenses entirely in Go, or entirely in CUE. We could relax this later.

This isn't fully ready - definitely needs more tests, and the requisite changes to `Translate()` aren't yet implemented. (Though i think those will be like another 40 LoC...trivial). But the test that exists so far at least verifies the correctness of the checker which ensures all and only the exact set of Go migrations are being provided.

The test also shows what actually writing Go lenses might plausibly look like - though **PLEASE NOTE** that when this is put to use in kindsys, we'll be putting in a wrapping layer so that the developer doesn't need to interact directly with `*thema.Instance`.

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -54,7 +54,7 @@ type lensID struct {
 	From, To SyntacticVersion
 }
 
-func lid(to, from SyntacticVersion) lensID {
+func lid(from, to SyntacticVersion) lensID {
 	return lensID{from, to}
 }
 
@@ -218,7 +218,7 @@ func (ml *maybeLineage) checkGoLensCompleteness() error {
 	// TODO(sdboyer) it'd be nice to consolidate all the errors so that the user always sees a complete set of problems
 	all := make(map[lensID]bool)
 	for _, lens := range ml.implens {
-		id := lid(lens.To, lens.From)
+		id := lid(lens.From, lens.To)
 		if all[id] {
 			return fmt.Errorf("duplicate Go migration %s", id)
 		}
@@ -234,7 +234,7 @@ func (ml *maybeLineage) checkGoLensCompleteness() error {
 	for _, sch := range ml.schlist[1:] {
 		// there must always at least be a reverse lens
 		v := sch.Version()
-		revid := lid(prior, v)
+		revid := lid(v, prior)
 
 		if !all[revid] {
 			missing = append(missing, revid)
@@ -244,7 +244,7 @@ func (ml *maybeLineage) checkGoLensCompleteness() error {
 
 		if v[0] != prior[0] {
 			// if we crossed a major version, there must also be a forward lens
-			fwdid := lid(v, prior)
+			fwdid := lid(prior, v)
 			if !all[fwdid] {
 				missing = append(missing, fwdid)
 			} else {
@@ -272,7 +272,7 @@ func (ml *maybeLineage) checkGoLensCompleteness() error {
 		// walk the slice so output is reliably ordered
 		for _, lens := range ml.implens {
 			// if it's not in the list it's because it was expected & already processed
-			elid := lid(lens.To, lens.From)
+			elid := lid(lens.From, lens.To)
 			if _, has := all[elid]; !has {
 				continue
 			}
@@ -301,7 +301,7 @@ func (ml *maybeLineage) checkGoLensCompleteness() error {
 
 	ml.lensmap = make(map[lensID]ImperativeLens, len(ml.implens))
 	for _, lens := range ml.implens {
-		ml.lensmap[lid(lens.To, lens.From)] = lens
+		ml.lensmap[lid(lens.From, lens.To)] = lens
 	}
 
 	return nil
--- a/bind.go
+++ b/bind.go
@@ -215,11 +215,15 @@ func (ml *maybeLineage) checkLensesOrder() error {
 }
 
 func (ml *maybeLineage) checkGoLensCompleteness() error {
+	// TODO(sdboyer) it'd be nice to consolidate all the errors so that the user always sees a complete set of problems
 	all := make(map[lensID]bool)
 	for _, lens := range ml.implens {
 		id := lid(lens.To, lens.From)
 		if all[id] {
 			return fmt.Errorf("duplicate Go migration %s", id)
+		}
+		if lens.Mapper == nil {
+			return fmt.Errorf("nil Go migration func for %s", id)
 		}
 		all[id] = true
 	}
--- a/bind.go
+++ b/bind.go
@@ -1,12 +1,14 @@
 package thema
 
 import (
+	"bytes"
 	"fmt"
 
 	"cuelang.org/go/cue"
 	cerrors "cuelang.org/go/cue/errors"
 	"cuelang.org/go/cue/token"
 	"github.com/cockroachdb/errors"
+
 	terrors "github.com/grafana/thema/errors"
 	"github.com/grafana/thema/internal/compat"
 )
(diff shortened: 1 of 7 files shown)


bind.go
```diff
@@ -189,6 +213,86 @@ func (ml *maybeLineage) checkLensesOrder() error {
...
+			if !all[fwdid] {
+				missing = append(missing, fwdid)
+			} else {
+				delete(all, fwdid)
+			}
+		}
+		prior = v
+	}
+
+	// TODO is it worth making each sub-item into its own error type?
```
joanlopez, line 248: Personally, I don't see any need (at least for now), to explicitly model every item into its own error, as far as we bring that information back to the consumer.

bind.go
```diff
@@ -189,6 +213,86 @@ func (ml *maybeLineage) checkLensesOrder() error {
...
+			if _, has := all[elid]; !has {
+				continue
+			}
+			if !synvExists(ml.allv, lens.To) {
+				fmt.Fprintf(b, "\t%s (schema version %s does not exist", elid, lens.To)
+			} else if !synvExists(ml.allv, lens.From) {
+				fmt.Fprintf(b, "\t%s (schema version %s does not exist", elid, lens.From)
+			} else if elid[0] == elid[1] {
+				fmt.Fprintf(b, "\t%s (self-migrations not allowed)", elid)
+			} else if elid[0].Less(elid[1]) {
```
joanlopez, line 276: Nit; I usually find a bit difficult to reason about _to->from->to_ with slice indexes (_[0], [1]..._), and even more considering I'm a LTR reader and in Thema we usually define lenses in the opposite order.

So, for the sake of making this cases a bit easier to follow and reason about, I think we could get benefit of the type introduce in this pull request `type lensID ...` to add some semantic methods for these checks (`elid[0] == elid[1]`, elid[0].Less(elid[1])`, etc).

Wdyt?

bind.go
```diff
@@ -179,7 +203,7 @@ func (ml *maybeLineage) checkLensesOrder() error {
 			return err
 		}
 
-		if err := checkLensesOrder(previous, &curr); err != nil {
+		if err := ml.doCheck(previous, &curr, ml.transplan); err != nil {
```
joanlopez, line 206: Looks like `ml.transplan` is never filled, which I guess it just corresponds to the note left on the description _`and the requisite changes to Translate() aren't yet implemented`_, so just a friendly reminder! :)

surface.go
```diff
@@ -185,6 +195,25 @@ func SkipBuggyChecks() BindOption {
 	}
 }
 
+// ImperativeLenses takes a slice of [ImperativeLens]. These lenses will be
+// executed on calls to [Instance.Translate].
+//
+// A mix of Go and CUE lenses may be provided, but only one lens may be provided
```
joanlopez, line 201: Note: we may need to revisit this text before getting this pull request merged - it doesn't seem to be aligned with what's stated in the pull request description.

bind.go
sdboyer: yeah, i was having a similar thought while writing this and getting lost in the logic. A couple things occurred to me:

* Make it a struct with explicit fields rather than positional in an array
* The typical way of printing this is LTR, being `<From> -> <To>` - perhaps that's also how the fields should be ordered

bind.go
sdboyer: yeah. `transplan` is actually from the first pass, where i was allowing a mix of CUE and Go. i hadn't cleaned it up yet or firmly decided whether or not to use it.

surface.go
```diff
@@ -185,6 +195,25 @@ func SkipBuggyChecks() BindOption {
...
+// across both languages for each necessary lens. If the set of lenses is
+// incomplete, or a duplicate lens is provided, [BindLineage] will fail.
+//
+// Lenses written in Go are Turing-complete, and therefore cannot be checked for
```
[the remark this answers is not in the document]
sdboyer: Agreed, testing is the proper way, i don't think we increase system correctness in a meaningful reliable way by trying to check anything with that at bind time

bind.go
```diff
@@ -39,10 +41,26 @@ type maybeLineage struct {
...
+
+	// translation plan, tracks whether each lens is handled in cue or go
+	transplan map[lensID]bool
+
 	// The raw input value is the root of a package instance
 	// rawIsPackage bool
 }
 
+// to, from
+type lensID [2]SyntacticVersion
```
DanCech, line 54: any reason not to use a struct here? seems like it could easily be confusing which is from and which is to

sdboyer: indeed! https://github.com/grafana/thema/pull/187#discussion_r1270338534

bind.go
```diff
@@ -39,10 +41,26 @@ type maybeLineage struct {
...
+
+func lid(to, from SyntacticVersion) lensID {
```
DanCech, line 56: similarly, reason for not defining this as from, to? This seems backward, I would say "I am going from `a` to `b`", not "I am going to `b` from `a`".

bind.go
DanCech: just came to this now after writing the same comments above

bind.go
sdboyer: Not a good one, no. it's all a weird artifact of probably mistakenly trying to be consistent with [the expected sort ordering in lenses](https://github.com/grafana/thema/blob/2a57611eed5ce3fcb08eef5c68cb4c05de2b131f/lineage.cue#L49-L75), where we sort first by `to` and then by `from`. (That declared sort ordering requirement only applies for CUE lenses, not for these Go things)

joanlopez: Just for curiosity; Could you give more context about the reasons behind such requirement, please? Thanks!

sdboyer: Requiring the inputs to be manually sorted by the user was the main way we got away from the need to dynamically sort them in CUE - which was the main source of our performance issues fixed in #173

joanlopez: Sorry, I mean specifically about the order: _first by `to` and then by `from`_

sdboyer: ahhh - i'm not attached to that as an ordering, either. My motivation was just that, assuming an ascending version sort order (same as with `schemas`), it seemed preferable to have forward lenses precede reverse ones in the list.

joanlopez (approved):

sdboyer:
i'm gonna merge these, with this note...

Recent discussions have absolutely led me to change my mind about the necessity of having Go lenses as an essential part of Thema. They just make all of what Thema is doing considerably easier to at least get started with, because it allows writing the more complex part (lenses) in a familiar language. All the checkable invariants in the world don't matter if the learning curve is too steep.

I'm not sure that what's in this PR is the final form that imperative lenses should take. It does seem to me like there's a plausible path where we actually handle lacuna resolvers as injected funcs, and that the implementation here could be wholly replaced by that. But this PR does the absolute minimal thing necessary to enable lenses in Go, which gets us started. And because it requires a clean split - either there are imperative lenses or declarative ones, no mixing and matching - we have the simplest foundation for changing it later and minimizing impact on users.

PR #188, opened 2023-07-21
Use _#schema to generate TS files

spinillos:
It rollback to the previous implementation since we need to use `_#schema`.

If you need some context: `_#schema` is a union between the `schema` and the `_join` value. We need this `_join` value to merge any extra information to the user's schemas (for example, validators).

For any reason, `Cuetsy` didn't like this union and it was unable to find the extends for the interfaces. We decided to split `_#schema` values temporally and merge them later to make it works. Finally we have a fix in `Cuetsy` to make it works.


sdboyer:
Glad we finally got this sorted!

Let's use this as an excuse to introduce some simple corpus tests for ts as well - same as we have for `encoding/gocode`.

encoding/typescript/gen.go
```diff
@@ -64,41 +64,38 @@ func GenerateTypes(sch thema.Schema, cfg *TypeConfig) (*ast.File, error) {
 	}
 	if cfg.CuetsyConfig == nil {
 		cfg.CuetsyConfig = &cuetsy.Config{
+			ImportMapper: func(_ string) (string, error) {
+				return "", nil
+			},
```
joanlopez, line 69:
```suggestion
			ImportMapper: cuetsy.IgnoreImportMapper,
```

joanlopez: I think this should work, it's the same, but [defined in Cuetsy](https://github.com/grafana/cuetsy/blob/main/import_mapper.go#L27C6-L27C24).

testdata/lineage/basic-multiversion.txtar
```diff
@@ -1840,3 +1840,59 @@ type Basicmultiversion struct {
 
 // BasicmultiversionWithDefault defines model for Basicmultiversion.WithDefault.
 type BasicmultiversionWithDefault string
+
+-- out/encoding/typescript/TestGenerate/nilcfg --
+export interface Basic-Multiversion {
```
joanlopez, line 1845: I don't think a Typescript identifier can contain `-`

encoding/typescript/gen.go
spinillos: I didn't recall that we had that!

testdata/lineage/basic-multiversion.txtar
spinillos: We didn't add anything in Cuetsy to normalise the fields and it should do it.

joanlopez: So, do you want me to create an issue in Cuetsy repository?

spinillos: I added them in Cuetsy and as-code

joanlopez: Thanks so much! :)

PR #189, opened 2023-07-21
Fixes test results

spinillos:
Tests were failing because they have invalid path in results 🤷🏼‍♀️


spinillos:
If I fix `lineage_test.go`, `vanilla_txtar.go` fails 😂: [drone link](https://github.com/grafana/thema/actions/runs/5621511614/job/15232432517?pr=189#step:4:420)

spinillos:
`go test -v ./...` works in local and CI, but `lineage_test.go` fails if you execute it independently

spinillos:
I'm going to close it for now. I created that only to fix that test and it should be different 😅.

PR #191, opened 2023-08-01
Expose Go post-processing utilities

K-Phoen:
This PR exposes the Go post-processing function, to be used by kindsys and other Go jennies.

### DIFF
--- a/encoding/gocode/gen.go
+++ b/encoding/gocode/gen.go
@@ -4,9 +4,6 @@ import (
 	"bytes"
 	"embed"
 	"fmt"
-	"go/parser"
-	"go/token"
-	"path/filepath"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


encoding/gocode/gen.go
```diff
@@ -123,10 +123,10 @@ func GenerateTypesOpenAPI(sch thema.Schema, cfg *TypeConfigOpenAPI) ([]byte, err
...
 
 	return PostprocessGoFile(GenGoFile{
-		Path:     fmt.Sprintf("%s_type_gen.go", sch.Lineage().Name()),
-		Appliers: applyFuncs,
-		In:       []byte(gostr),
-		ErrIfAdd: !cfg.IgnoreDiscoveredImports,
+		Path:                    fmt.Sprintf("%s_type_gen.go", sch.Lineage().Name()),
+		Appliers:                applyFuncs,
+		In:                      []byte(gostr),
+		IgnoreDiscoveredImports: !cfg.IgnoreDiscoveredImports,
```
joanlopez, line 129: Doesn't it look weird that `IgnoreDiscoveredImports: !IgnoreDiscoveredImports`? 🤔 

Also, based on the underlying behavior, I'm not sure if the name is clear enough. I guess nothing that cannot be clarified with some extra explanation in form of docs, if needed.

K-Phoen: That's a copy-pasta mistake, good catch!

K-Phoen:
Fixed, documented and added a small test case for it :)

PR #192, opened 2023-08-01
Simplify the regex used to trigger the Go github action on tags

K-Phoen:
The previous regex wasn't accepted by the GitHub actions runner (see https://github.com/grafana/thema/pull/180#issuecomment-1658189427).

The following change should allow the action to run again, at the cost of a less constraining verification made on the tags themselves.

My goal here is to get a CI back up and running quickly, improvements to the regex can be made later if/when needed.


K-Phoen:
Also: to avoid the noise due to a failing "goreleaser" action in PRs, I split the release and CI actions.

sdboyer:
thank you!

